PR #1, opened 2015-08-28
Provide ESRI and EPSG codes

yeesian:
and with common spatial references provided.

I fully understand if *Fugro Roames* decides to maintain their own package, but am helping out (given that it's under the MIT license) under the `LibPROJ.jl` fork, with attributions where they are due.


c42f:
Thanks, this is great, particularly the big list of epsg codes.  I've every intention of merging changes from outside, it would be pointless to announce an "open source" package and only accept patches from internal contributors ;-)

I'll look in more detail later, but would you consider making/discussing the rename in a separate pull request or on the mailing list?  I'm not yet quite convinced that LibPROJ is the best name.  If nothing else it's not the accepted standard in the wider julia community to name library wrappers as LibBlah - just look at the list of registered packages in METADATA.jl.

c42f:
Sorry to be pedantic about naming, but I find the choice of name to be an oddly personal thing.  I do want opinions, but I don't want to be rushed into it ;-)

yeesian:
Yeah sure, don't worry about it! I'm okay with sticking to the `Proj4` name. I was more worried about giving the impression of pressuring you into transferring the ownership to JuliaGeo, if Fugro Roames prefers to maintain their own fork, and was prepared to maintain my own separate fork (under a different name) if need be.

Let me know how you prefer it to work out -- I can push a separate commit to this PR to rename it back, if need be!

c42f:
Cheers, thanks.  There should certainly be a single definitive copy of this (that is, the place that METADATA.jl points to) - having two disparate forks would just be pointlessly confusing to julia geo users.  As I said, I'd be happy to make this part of JuliaGeo in principle.  There's just some things we need to work out internally about how to promote our open source work so everyone at Fugro Roames is happy with the arrangement.

I do think these goals are mutually compatible.  I hope you can put up with merging your changes through here for the meantime?

Regarding the rename - yes please just rebase the rename changes away, or revert them in a separate commit.  When a new name has been decided, it's easy to put them back in.

yeesian:
> I hope you can put up with merging your changes through here for the meantime?

Sure!

src/Proj4.jl
```diff
@@ -50,14 +48,21 @@ function _get_def(proj::Projection)
...
 end
 
-
+function _transform!(src::Projection, dest::Projection, point_count::Int64,
+                     point_stride::Int32, x::Ptr{Float64}, y::Ptr{Float64})
+    @assert src.rep != C_NULL && dest.rep != C_NULL
+    ccall((:pj_transform, libproj), Cint,
+          (Ptr{Void}, Ptr{Void}, Clong, Cint, Ptr{Cdouble}, Ptr{Cdouble}, Ptr{Void}),
+          src.rep, dest.rep, point_count, point_stride, x, y, C_NULL)
+end
```
c42f: Why is this extra version of `_transform!` necessary?  It seems like seven added lines for no added functionality over what was there before?

src/Proj4.jl
```diff
@@ -50,14 +48,21 @@ function _get_def(proj::Projection)
 end
 
 """Low level interface to libproj transform, allowing user to specify strides"""
-function _transform!(src::Projection, dest::Projection, point_count, point_stride, x, y, z)
+function _transform!(src::Projection, dest::Projection, point_count::Int64,
+                     point_stride::Int32, x::Ptr{Float64}, y::Ptr{Float64}, z::Ptr{Float64})
```
c42f: Not specifying the types for `point_count` and `point_stride` was intentional here - as I understand it, `ccall` will convert these to the necessary C types internally, so I thought keeping it generic made some kind of sense: it's perfectly fine to call the function with a `point_count` of Int32 for example.

Actually, the underlying C interface uses `Cint` and `Clong`, which may very well both be `Int32`, even on a 64 bit machine, depending on the platform ABI.

src/Proj4.jl
yeesian: This is for dealing with calls to transforms for 2-dimensional points, without having to pass in `C_NULL` for z (otherwise it will be unclear whether it is a `Ptr{Float64}` or a `Ptr{Void}`)

src/Proj4.jl
```diff
@@ -125,8 +130,9 @@ function transform!(src::Projection, dest::Projection, position::Array{Float64,2
     P = pointer(position)
     x = P
     y = P + sizeof(Cdouble)*npoints
-    z = ncomps < 3 ? C_NULL : P + 2*sizeof(Cdouble)*npoints
-    err = _transform!(src, dest, npoints, 1, x, y, z)
+    z = P + 2*sizeof(Cdouble)*npoints
+    err = (ncomps < 3) ? _transform!(src, dest, npoints, Int32(1), x, y) :
+          _transform!(src, dest, npoints, Int32(1), x, y, z)
```
c42f: Again, as above - the version with less lines makes more sense to me, since it seems to have the same functionality.  I'm certainly happy to go with the change if there's a good reason for it ;-)

src/Proj4.jl
```diff
@@ -136,6 +142,9 @@ function transform!(src::Projection, dest::Projection, position::Array{Float64,2
     position
 end
 
+transform!(src::Projection, dest::Projection, position::Vector{Float64}; radians::Bool=false) =
+    transform!(src, dest, reshape(position,(1,length(position))), radians=radians)
+
```
c42f: Looks good.

General question: do you think specifying sets of N coordinates as Nx3 matrices makes sense in the context of other julia libraries you've used, vs the alternative of 3xN?  The option here is the way I've habitually done it in matlab, so that's what I've got here.  However, some libraries (eg, KDTrees) want the transpose which is a pain.

src/Proj4.jl
yeesian: You're right that I haven't been very consistent in sticking to the platform independent types

src/Proj4.jl
yeesian: From what I've seen so far, usually (vector) geographical data doesn't come as nice contiguous matrices, and are more commonly represented as (nested) vector-of-vectors

yeesian: The reasoning (from what I understand) is that usually parsers/developers cannot determine the sizes the linestring/polygon in advance

src/Proj4.jl
```diff
@@ -154,18 +163,23 @@ Returns:
...
-end
 
-function transform{T}(src::Projection, dest::Projection, position::Array{T,2}; radians::Bool=false)
-    transform(src, dest, map(Float64, position), radians=radians)
-end
+transform(src::Projection, dest::Projection, position::Vector{Float64}; radians::Bool=false) =
+    transform!(src, dest, copy(reshape(position,(1,length(position)))), radians=radians)
 
+transform{T <: Real}(src::Projection, dest::Projection, position::Array{T,2}; radians::Bool=false) =
+    transform!(src, dest, map(Float64,position), radians=radians)
```
c42f: Right, I think it makes sense to restrict `T` as `Real` here.

src/README.md
```diff
@@ -0,0 +1,11 @@
+# Spatial Reference Objects
+
+A Spatial Reference System Identifier (SRID) is a unique value used to unambiguously identify projected, unprojected, and local spatial coordinate system definitions. Each projected and geographic coordinate system is defined by either a well-known ID (WKID) or a definition string (WKT).
+
+Virtually all major spatial vendors have created their own SRID implementation or refer to those of an authority, such as the European Petroleum Survey Group (EPSG), or Environmental Systems Research Institute (ESRI).
+
+We provide projection codes through their WKIDs in the `epsg` and `esri` dictionaries in `src/projection_codes.jl`. For the uninitiated, their distinction is stated in [the following response](http://gis.stackexchange.com/a/18675) from Michael D. Kennedy:
```
c42f: I think it would be worth noting that these are taken directly from the underlying proj.4 projection databases, just to be clear.

src/Proj4.jl
yeesian: In the original version, the type of `z` varies between `Ptr{Float64}` and `Ptr{Void}`, depending on the value of `ncomps`.

Ideally, it'd be better to determine from the data-type of `z` which version of `transform!` to dispatch it to, rather than based on the run-time `length`, but since we've decided to be relaxed about supporting vectors (which can be arbitrary in length), we have to perform these runtime checks.

src/README.md
yeesian: I have tried to stay as close to the original source of the projection strings as possible, but I think we'll have to be conscientious about updating them whether it changes upstream. To that end, I've (i) added some tests for them, and (ii) tracked the date/time when I change the esri/epsg codes

test/runtests.jl
```diff
@@ -1,11 +1,104 @@
...
+
+# lla_point = [1.290270 103.851959 0.0]
+
+for proj_string in values(Proj4.epsg)
+   proj = Projection(proj_string)
+   proj_string1 = Proj4._get_def(proj)
+   proj1 = Projection(proj_string1)
+   proj_string2 = Proj4._get_def(proj1)
+   proj2 = Projection(proj_string2)
+   proj_string3 = Proj4._get_def(proj2)
```
c42f: It's great to have some proper testing, thanks.

Having to use the internal `_get_def()` this way makes me think that there should be an public exported function which does the same thing, or at least something with a nicer name.

It's not quite clear to me which function in base should be overloaded, if any.  As you can see, I've overloaded `show()`, but that adds the extra `Projection()` wrapper around the string.  There's a bunch of somewhat likely looking functions... `repr()`, `string()`, `show()` etc, all of which ultimately call `show()` at the moment.

src/README.md
c42f: Or should we make generating the strings part of the build process?  It would be easy to put your generation script into build.jl instead.

c42f: Including downloading them from the original source - I don't like the idea of having to remember to update this stuff.

yeesian: That might make it different for different users though.. I'll rather we tag a new release everytime we update the projection strings, to make it easier to track changes?

yeesian: We don't have to remind ourselves to update this stuff periodically; just need to wait for people to file issues =p

src/Proj4.jl
c42f: Shouldn't this be resolved by relaxing the type constraints, and let `ccall` check for type compatibility instead.  After all, `_transform!` is meant to be an internal function.  If that's not to your liking, you can just make `z` into `Union(Ptr{Float64},Ptr{Void})` ?

src/Proj4.jl
c42f: So the worry here is type instability?

src/Proj4.jl
c42f: Nested vectors? Ick.

At work we deal with a lot of point clouds (not polygons), so the data is usually nicely structured.

src/Proj4.jl
yeesian: Yes, although we aren't exactly benefitting from the change as of now (yet).

src/README.md
c42f: Heh, ok I'm happy to leave the build as-is.

I still think it would be worth noting the original source for the projection definitions in the documentation.

src/Proj4.jl
c42f: In any case, let's leave the matrices as Nx3 for now.

yeesian: :ok_hand:

src/Proj4.jl
c42f: There's a very simple solution for type stability: just use `Ptr{Float64}(C_NULL)`.  Right?

c42f: that is,

```julia
z = ncomps < 3 ? Ptr{Float64}(C_NULL) : P + 2*sizeof(Float64)*npoints
```

yeesian: I was not aware of it, thanks! Seems to work only for julia v0.4 though, But you're right, I've removed the redundant method

src/Proj4.jl
```diff
@@ -34,7 +34,7 @@ end
 
 "Get human readable error string from proj.4 error code"
 function _strerrno(code::Cint)
-    bytestring(ccall((:pj_strerrno, libproj), Cstring, (Cint,), code))
+    bytestring(ccall((:pj_strerrno, libproj), Ptr{Cchar}, (Cint,), code))
```
c42f: Good work with the v0.3 stuff.

I'd suggest to leave this as `Cstring` - that's available in `Compat.jl`, or am I mistaken?

src/Proj4.jl
c42f: Cheers, looks good.

src/Proj4.jl
yeesian: Unfortunately didn't work :/

```julia
julia> Compat.@compat(Cstring)
ERROR: Cstring not defined
```

src/Proj4.jl
yeesian: Doesn't work for v0.3 though, haha

src/Proj4.jl
c42f: I think that's because it's not defined via the `@compat` macro - it's just a constant type alias.  See
https://github.com/JuliaLang/Compat.jl/blob/bc9f16c6e582345df8517d4f0708fff8138c2baa/src/Compat.jl#L529

src/Proj4.jl
```diff
@@ -34,7 +34,7 @@ end
 
 "Get human readable error string from proj.4 error code"
 function _strerrno(code::Cint)
-    bytestring(ccall((:pj_strerrno, libproj), Cstring, (Cint,), code))
+    bytestring(ccall((:pj_strerrno, libproj), Compat.Cstring, (Cint,), code))
```
c42f: Almost!  This won't work on julia 0.4 though :-/  I think you simply want to use `Cstring` with no decoration - `using Compat` should bring the declaration into scope correctly.

src/Proj4.jl
```diff
@@ -46,19 +46,18 @@ end
...
     @assert proj.rep != C_NULL
     opts = 0 # Apparently obsolete argument, not used in current proj source
-    bytestring(ccall((:pj_get_def, libproj), Cstring, (Ptr{Void}, Cint), proj.rep, opts))
+    bytestring(ccall((:pj_get_def, libproj), Compat.Cstring, (Ptr{Void}, Cint), proj.rep, opts))
 end
 
 """Low level interface to libproj transform, allowing user to specify strides"""
-function _transform!(src::Projection, dest::Projection, point_count, point_stride, x, y, z)
+function _transform!(src::Projection, dest::Projection, point_count::Clong,
+                     point_stride::Cint, x::Ptr{Cdouble}, y::Ptr{Cdouble}, z::Ptr{Cdouble})
```
c42f: I'm still not sure that writing the C types is really necessary here for `point_count` and `point_stride`.  To me it seems just to be forcing you to do extra type casting down below.  According to the `ccall` documentation, there's an implicit call to `cconvert()` inserted for each argument passed to the `ccall` function - that would seem to completely take care of type conversion worries.

test/runtests.jl
c42f: After some reading, it looks like the correct thing to do is to overload `Base.print()` for the undecorated version, keeping `Base.show()` as already defined above.  Having done that, we should be able to replace `Proj4._get_def(proj)` with a simple call to `string(proj)` here.

c42f: (I pushed a small change to master to implement `print()` and shorten `show()`)

src/Proj4.jl
```diff
@@ -1,6 +1,8 @@
 module Proj4
 
-export transform, transform!, Projection, is_latlong, is_gencent
+import Compat
```
c42f: Ah, I see the problem you're having below - this should be `using Compat` instead.

src/Proj4.jl
```diff
@@ -101,7 +100,7 @@ end
 """
 Return true if the projection is a geocentric coordinate system
 """
-function is_gencent(proj::Projection)
+function is_geocent(proj::Projection)
```
c42f: Oops, good catch.

src/Proj4.jl
```diff
@@ -125,8 +124,8 @@ function transform!(src::Projection, dest::Projection, position::Array{Float64,2
     P = pointer(position)
     x = P
     y = P + sizeof(Cdouble)*npoints
-    z = ncomps < 3 ? C_NULL : P + 2*sizeof(Cdouble)*npoints
-    err = _transform!(src, dest, npoints, 1, x, y, z)
+    z = (ncomps < 3) ? Ptr{Float64}(C_NULL) : P + 2*sizeof(Cdouble)*npoints
+    err = _transform!(src, dest, npoints, Compat.@compat(Int32(1)), x, y, z)
```
c42f: If you avoid overspecifying the types above, you can avoid this cast.  I think...

test/runtests.jl
yeesian: It's common that people will want to get the projection string from a Projection. I'll push a `proj4string` method (name taken from python's basemap: https://github.com/matplotlib/basemap/search?q=proj4string). See if you like it.

src/Proj4.jl
yeesian: I'll revert it back to what you have originally then. Sorry for the churn!

src/Proj4.jl
yeesian: :ok_hand:

yeesian:
Okay, I think I'm good here -- ready for your review. Thanks for taking your time to go through it!

c42f:
Cool, I pulled this and played around with it a bit.

With the latest version I get a failure in the tests: "LoadError: Could not parse projection string: "+proj=longlat +datum=carthage +no_defs": unknown elliptical parameter name".  This is because we're pulling the proj strings directly from the latest proj.4 master, rather than using the ones which come with the installed version of the library - if I build the latest libproj from git it doesn't seem to be an issue.  However, I don't really want to force users to have the latest version of a very stable library like proj, they should be able to use the version from their package manager.

Ok, so that's an annoying complication, and it suggests going back to the previous idea of building the proj dictionaries using build.jl.  If we did that, the actual data would need to be fetched from the package manager version.  On Ubuntu, that's the proj-data package.  It's a nasty complication, I'll give you that.

An additional problem is load time has skyrocketed with loading the huge dictionaries of transformations.  Perhaps we really shouldn't have these here at all, and instead make a new package ProjData.jl or some such?  Another option would be to load them dynamically from a Proj4.jl data path.

src/Proj4.jl
```diff
@@ -101,11 +99,14 @@ end
...
 Return true if the projection is a geocentric coordinate system
 """
-function is_gencent(proj::Projection)
+function is_geocent(proj::Projection)
     @assert proj.rep != C_NULL
     ccall((:pj_is_geocent, libproj), Cint, (Ptr{Void},), proj.rep) != 0
 end
 
+"Get projection definition string in the proj.4 plus format"
+proj4string(proj::Projection) = _get_def(proj)
```
c42f: Did you consider the version I pushed to master before I raised this yesterday?  In that, I just overloaded `Base.print(::IO,::Projection)`, which seems like it might be the right thing (though I'm not completely certain).  I had to dig surprisingly deeply into the documentation to work out precisely which functions seem like the right ones to use.  Here's the reference for the distinction between `show()` and `print()`.

http://julia.readthedocs.org/en/latest/stdlib/io-network/#Base.print
http://julia.readthedocs.org/en/latest/stdlib/io-network/#Base.show

c42f: (Having overloaded `print()` in this way, `string()` automatically works.)

yeesian:
> With the latest version I get a failure in the tests: "LoadError: Could not parse projection string: "+proj=longlat +datum=carthage +no_defs": unknown elliptical parameter name".

Sounds like a feature and not a bug. I'm glad they are at least pointing the user to the fact that their current binary doesn't support a given projection string, and they'd need to either (i) update their libproj binary, or (ii) not be able to parse the projection string anyway.

That said, I/we should refactor the tests such that it proceeds with the rest of the tests, rather than just erroring out.

> An additional problem is load time has skyrocketed with loading the huge dictionaries of transformations. Perhaps we really shouldn't have these here at all, and instead make a new package ProjData.jl or some such? Another option would be to load them dynamically from a Proj4.jl data path.

I wouldn't worry about load times, given that we can now precompile packages in Julia.

src/Proj4.jl
yeesian: oh! that's cool :D I'll remove this line then (and the corresponding calls in the tests), and leave you to it when you merge?

yeesian:
I dug into the PROJ.4 repo today, and noticed they have introduced an [API for performing geodesic operations](https://github.com/OSGeo/proj.4/blob/master/src/geodesic.h), but am holding off on pushing more changes to this PR.

At this point I have just a few things in mind, that I hope to accomplish over the weekend:
- [ ] code reorganization.
- [ ] introduce geodesic operations (see https://github.com/OSGeo/proj.4/issues/34)
- [ ] add documentation/tests

Will you be looking into providing a build script? Feel free to have a look at [my attempt for LibGEOS](https://github.com/JuliaGeo/LibGEOS.jl/pull/7).

What do you think is the best way to proceed?

c42f:
Yes, let's not make any more changes here (other than fixes) - we've already got the dictionaries and various 0.3 compatibility stuff munged together.

If you're happy with it, I'd prefer to squash the changes herein before I push, since there has been some code churn.  Obviously I'd make sure the squashed commit is properly attributed to you.

Regarding broken tests - I'm quite sure this is not a feature, it's a bug: while it's true that the underlying proj library may not be able to parse all the strings, this is not relevant to the functionality or correctness of the *wrapper* code.  We can report on which projections couldn't be parsed by the underlying library, but this is IMO not a reason to fail the tests.

src/Proj4.jl
c42f: Yes, no problems

yeesian:
Okay squashed and rebased. I've removed the commit where I introduced the `proj4string` method. Can you help check on whatever versions of julia you might have before merging?

c42f:
Oh nice, this is much neater thanks.

yeesian:
> Regarding broken tests - I'm quite sure this is not a feature, it's a bug: while it's true that the underlying proj library may not be able to parse all the strings, this is not relevant to the functionality or correctness of the wrapper code. We can report on which projections couldn't be parsed by the underlying library, but this is IMO not a reason to fail the tests.

Yeah, I think we can change to printing out the projection strings that couldn't be parsed, instead of having them as tests. Do I do that in this PR?

c42f:
> Yeah, I think we can change to printing out the projection strings that couldn't be parsed, instead of having them as tests. Do I do that in this PR?

Sure, it's definitely relevant to this PR.  Regarding next steps, I'll make an issue where we can discuss steps to get to a registered version of the package.

c42f:
I made a todo list for 0.1 at #2.  Discussion welcome!

c42f:
Ok, I made a couple of minor adjustments and merged this.

On closer inspection I'm not quite sure why you need a three way round trip with the projection string tests, but we can clean that up later if it makes sense.

PR #4, opened 2015-08-31
Code Reorganization and Geodesic Routines

yeesian:
Here's some collected thoughts after having worked on it over the weekend,

# Code Organization
We should separate the low-level C-facing routines from the high-level user-facing ones.

## lower-level
In the low-level C-facing routines (`src/geodesic.jl` and `src/proj_capi.jl`), that's where

- pointer arithmetic should be handled
- error messages should be dealt with
- C-type immutables/structs (for the express purpose of interfacing with the PROJ.4 library, rather than for users) should be defined, and dealt with locally

Preferably the only function arguments that should be pointers are native objects from the proj.4 library. I've made an exception for one of the `_transform` methods, because it gets re-used alot in the other calls to `transform` (for reshaping between `Array{T,2}` and `Vector{T}`).

## higher-level
The higher-level user-facing routines are organized into `src/proj_types.jl` and `src/proj_functions.jl`. This being a wrapper library, I'm biased towards
- ordered arguments, rather than keyword arguments.
- concretely-typed variables, rather than loosely-typed ones
- surfacing decisions for the user (through function arguments/naming), rather than making decisions for them; convenience functions can be added at a later stage

This is most obviously felt in `transform`, `transform2` and `transform3` (both with and without the bang). It might feel like code-bloat, but the wider variety of function calls is actually useful for the end-user: most of the time, they would know whether or not the data they have is 2 or 3 dimensional. In such cases, they will prefer to save on the checking/branching by calling the most specific function, especially if they're doing the geodesic operations in a tight loop.

# Support
We should also provide consistency in user expectation (if violated, it's always from my own wrongdoing), rather than simply documenting interfaces:

   1. calls to functions with vector arguments return vector results, and calls with NxM arrays return NxM results.
   2. Results should not end up mysteriously transposed/etc.
   3. If a argument of a certain type is allowed in some functions, it should be allowed in all "similar"/reasonable functions

(This has implications in the "other thoughts section" when it comes to tightening the range of allowable argument types.)

I don't mind if someone else comes along, and submits a PR to support v0.3, but I personally don't have the energy to maintain that level of support, sorry! I have gone back to clean up all the unused functions (`geodesic2geocentric`, `latlong_from_projection`, etc) to scope it down into something I can manage. Also, I suggest we stick with Float64/EPSG/ESRI, and support more general user arguments (Int/Real/etc) only when we have the resources (and are in a good position) to. I personally think that time/effort is better spent on developing the rest of the missing functionality for working with geospatial data (e.g. OGR/etc) though.

# Other thoughts
- The semantics of `Array{T,2}` is not entirely clear-cut to me; I can imagine, for instance, that the columns (rather than the rows) might correspond to distinct points. Although it might feel more "general"(i.e. a vector can be "reshaped" into a 2-d matrix), it introduces a lot of degrees of freedom in how it might/should be interpreted. Moreover, it is weirder to support Arrays without also supporting Vectors, rather than vice versa
- The length of the code is secondary to me for now; it can be shortened via parameterizing the functions, or meta-programming at a later stage
- I have been rather wild and loose with my individual commits, since there was (and still is) a lot of ground to cover. I expect to change after we tag the first release, and hope it doesn't annoy you too much.
- re: https://github.com/FugroRoames/Proj4.jl/issues/2#issuecomment-135942011, I work from a macbook, so I can help with building/testing for OS X, but I hope you can figure out how the build service works too, to increase the bus factor of this package.


yeesian:
Alright, I'm done here (for now). Have at it!

src/proj_functions.jl
```diff
@@ -0,0 +1,211 @@
+@doc "Return true if the projection is a geographic coordinate system" ->
+is_latlong(proj::Projection) = _is_latlong(proj.rep)
+
+@doc "Return true if the projection is a geocentric coordinate system" ->
+is_geocent(proj::Projection) = _is_geocent(proj.rep)
+
+@doc "Return true if the projection is a geocentric coordinate system" ->
+is_identical(p1::Projection, p2::Projection) = _compare_datums(p1.rep, p2.rep)
```
c42f: I'm not an expert on this stuff, but my impression is that `pj_compare_datum` only compares a very specific part of the projection definition: the "datum" ie, the underlying latlong coordinate system.

Unfortunately proj doesn't seem to provide a way to fully compare projections - see the comment at the top of `pj_transform`: https://github.com/OSGeo/proj.4/blob/master/src/pj_transform.c#L77

c42f: Thus, I guess there's no neat way to implement `is_identical`.

yeesian: oh sure, i should stick to the original name `compare_datums` then. Copied wrongly from https://github.com/OSGeo/proj.4/blob/5d2af3a89be0898458e4f0fe272affc36642d304/src/pj_transform.c#L448-L453

yeesian:
For reference: https://github.com/jswhit/pyproj/blob/master/_proj.pyx

src/Proj4.jl
```diff
@@ -1,171 +1,23 @@
...
-    transform!(src, dest, copy(position), radians=radians)
-
-transform(src::Projection, dest::Projection, position::Vector{Float64}; radians::Bool=false) =
-    transform!(src, dest, copy(reshape(position,(1,length(position)))), radians=radians)
-
-transform{T <: Real}(src::Projection, dest::Projection, position::Array{T,2}; radians::Bool=false) =
-    transform!(src, dest, @compat(map(Float64,position)), radians=radians)
+export Projection, # proj_types.jl
+       transform2, transform2!, transform3, transform3!, transform, transform!,  # proj_functions.jl
+       is_latlong, is_geocent, is_identical, spheroid_params,
```
c42f: is_identical -> compare_datums now that you changed it.  Ditto for the tests.

c42f:
Right, I've only just had the chance to get back to this.  Some thoughts to start with:

The low level / high level split is something I had already started on in a minor way and I think it makes sense to move into separate files as you've done here.


> ## higher-level
> The higher-level user-facing routines are organized into `src/proj_types.jl` and `src/proj_functions.jl`. This being a wrapper library, I'm biased towards
> - ordered arguments, rather than keyword arguments.

I'm naturally biased toward readability, and I think keyword arguments are far better for "configuration" type stuff like the radians argument, where the usage pattern is very likely to involve a literal constant value in the calling code.

```julia
# Consider the positional argument version where a reader will need to look up
# the Pro4.jl API:
transform(srcCoords, destCoords, position, true)

# vs the entirely self-documenting alternative:
transform(srcCoords, destCoords, position, radians=true)
```

Now, I haven't measured the performance impact of doing this, and that may be a good counteragument.  In *principle* it's zero, but that doesn't seem to currently be the case.  I looked at the code_native output for a simple synthetic keyword example and it was a bit of a complexity disaster compared to the positional argument version.

(Example in question:
```julia
foo(x, s=false) = s ? x*x : x
bar(x) = foo(x, true)
code_native(bar, (Int,))  # super neat

foo2(x; s::Bool=false) = s ? x*x : x
bar2(x) = foo2(x, s=true)
code_native(bar2, (Int,)) # oh no what
```
)

> - concretely-typed variables, rather than loosely-typed ones

Concrete typing makes sense when it's required for passing reference to the underlying C API.  Clearly there's no point defining a `transform!` with anything other than `Cdouble` arrays, since the underlying API truly doesn't support anything else.  OTOH, what exactly does tight typing buy us in the case of value types?  From a user point of view it seems the difference between just working and tediously having to cast types?

> - surfacing decisions for the user (through function arguments/naming), rather than making decisions for them; convenience functions can be added at a later stage

I agree in the cases where the decision may affect performance, and where
there's no clearly correct API choice.

> This is most obviously felt in `transform`, `transform2` and `transform3` (both with and without the bang). It might feel like code-bloat, but the wider variety of function calls is actually useful for the end-user: most of the time, they would know whether or not the data they have is 2 or 3 dimensional. In such cases, they will prefer to save on the checking/branching by calling the most specific function, especially if they're doing the geodesic operations in a tight loop.

I don't really buy the argument that checking for dimensionality inside `transform()` - or other similarly simple operations - could be significant for performance.  Even simple cartographic projections are likely to involve quite a bit of floating point math internally which should swamp a simple and predictable branch instruction.  Lucikly, it should be easy to benchmark and actually find out :-)

src/Proj4.jl
yeesian: Sorry about that!

c42f:
Another note regarding the radians argument - this is entirely a convenience feature which I stole from pyproj.  Under the premise of providing convenience functions at a later stage, it should just be removed entirely, and people will have to stick with radians.

Not that I'm necessarily arguing for removing it - I  put in the API because I thought the pyproj people had made a good call about the need for the option, and the appropriate default, given that the world is basically stuck with degrees rather than following the one true way ;-)

yeesian:
Luckily for keyword args, you can define a new function:
```julia
transform(srcCoords, destCoords, position; radians=true) = transform(srcCoords, destCoords, position, radians)
```

c42f:
Good enough I guess, though a bit featury having so many versions of the same thing.

yeesian:
> OTOH, what exactly does tight typing buy us in the case of value types? From a user point of view it seems the difference between just working and tediously having to cast types?

Coming form python, I agree with you too, about how tedious it is, and it is definitely felt when writing tests. Some thoughts:

1. If it's for a casual end-user, playing around and exploring the code, it is mildly annoying
2. If it's for another program, it could range from ("gosh, so annoying") to ("okay, good to know"). I don't have the bandwidth to benchmark, but it makes sense to perform all the operations for conversions in batch, before performing all the projection operations
3. We can define higher-order functions that perform the conversions. I just elected to have not done it yet (the whole point of multi-dispatch was to have different behavior for different types; for wrapper libraries, not all types are created equal)
4. It would involve more work than I am willing to support in a single PR (c.f. 3) as, you have said it too, it feels featurey

c42f:
On to the other points.

> # Support
> We should also provide consistency in user expectation (if violated, it's always from my own wrongdoing), rather than simply documenting interfaces:

Yes, API designers sometimes blame users for not reading documentation, but I think that's often just as much lazy design as lazy users.  If a design is really done well, users should be able to guess, and guess correctly :)

>    1. calls to functions with vector arguments return vector results, and calls with NxM arrays return NxM results.

Agreed.

>    2. Results should not end up mysteriously transposed/etc.

Double agreed.  This is something that matlab APIs suffer from terribly, the main reason being that they try to guess intended meaning based on matrix shape.  Invariably this works in prototyping, only to fail horribly on an edge case in production code.

>    3. If a argument of a certain type is allowed in some functions, it should be allowed in all "similar"/reasonable functions

Sounds reasonable.

> (This has implications in the "other thoughts section" when it comes to tightening the range of allowable argument types.)
> 
> I don't mind if someone else comes along, and submits a PR to support v0.3, but I personally don't have the energy to maintain that level of support, sorry! I have gone back to clean up all the unused functions (`geodesic2geocentric`, `latlong_from_projection`, etc) to scope it down into something I can manage. Also, I suggest we stick with Float64/EPSG/ESRI, and support more general user arguments (Int/Real/etc) only when we have the resources (and are in a good position) to. I personally think that time/effort is better spent on developing the rest of the missing functionality for working with geospatial data (e.g. OGR/etc) though.

I never had any particular intention of supporting 0.3, at least not to start with.  I'm certainly not against it, but it's just not a priority for me with 0.4 so close.  I'm perfectly happy if you just rip out 0.3 support if you like.

> # Other thoughts
> - The semantics of `Array{T,2}` is not entirely clear-cut to me; I can imagine, for instance, that the columns (rather than the rows) might correspond to distinct points. Although it might feel more "general"(i.e. a vector can be "reshaped" into a 2-d matrix), it introduces a lot of degrees of freedom in how it might/should be interpreted. Moreover, it is weirder to support Arrays without also supporting Vectors, rather than vice versa

Indeed, transforming a single vector should be the base functionality, I only started with matrices because large sets of points are my use case.  The data layout (Nx3 vs 3xN) is a particularly nasty decision because both conventions are used in practice.

To make matters worse, some people may want to represent their data as vectors of ImmutableArrays.Vector3, or whatever the successor to that library will be once a more efficient NTuple appears in 0.5.  In terms of data layout, this is compatible with a 3xN matrix, but the user will be stuck with reinterpreting the data by hand.

I'm afraid this might be a hopeless case until some strong consensus arises in the julia community around the representation for small fixed size vectors.

> - The length of the code is secondary to me for now; it can be shortened via parameterizing the functions, or meta-programming at a later stage
> - I have been rather wild and loose with my individual commits, since there was (and still is) a lot of ground to cover. I expect to change after we tag the first release, and hope it doesn't annoy you too much.

No major problem, I'm reviewing all commits here as a big chunk.  If the individual commits themselves are messy and not very meaningful, you can always squash them together to neaten things up.

> - re: https://github.com/FugroRoames/Proj4.jl/issues/2#issuecomment-135942011, I work from a macbook, so I can help with building/testing for OS X, but I hope you can figure out how the build service works too, to increase the bus factor of this package.

I'll try to tackle this when I get time.  No guarantees this will be really soon - I expect real life to intrude in the shape of a new baby within a couple of weeks or so!

src/proj_functions.jl
```diff
@@ -0,0 +1,211 @@
...
+    transform3!(src, dest, copy(position[:,1:3]), radians)
+
+function transform3!(src::Projection, dest::Projection, position::Vector{Float64}, radians::Bool=false)
+    !radians && is_latlong(src) && (position[1:3] = deg2rad(position))
+    _transform3!(src.rep, dest.rep, position)
+    !radians && is_latlong(dest) && (position[1:3] = rad2deg(position))
+    position
+end
+transform3(src::Projection, dest::Projection, position::Vector{Float64}, radians::Bool=false) =
+    transform3!(src, dest, copy(position), radians)
```
c42f: Personally I don't much care for the explosion of different versions of `transform` here.  Is there a concrete reason that we really need all these?

src/proj_functions.jl
```diff
@@ -0,0 +1,211 @@
...
+    !radians && is_latlong(dest) && (position[1:2] = rad2deg(position))
+    position
+end
+transform2(src::Projection, dest::Projection, position::Vector{Float64}, radians::Bool=false) =
+    transform2!(src, dest, copy(position), radians)
+
+function transform3!(src::Projection, dest::Projection, position::Array{Float64,2}, radians::Bool=false)
+    !radians && is_latlong(src) && (position[:,1:3] = deg2rad(position[:,1:3]))
+    _transform3!(src.rep, dest.rep, position[:,1:3])
+    !radians && is_latlong(dest) && (position[:,1:3] = rad2deg(position[:,1:3]))
```
c42f: I think this accidentally converts the height component to and from radians - should be position[:,1:2] right?

src/proj_functions.jl
c42f: To be precise: I'm not seeing the value provided by transform{2,3} over plain transform().

src/proj_types.jl
```diff
@@ -0,0 +1,50 @@
...
+    rep::Ptr{Void} # Pointer to internal projPJ struct
+    
+    # [geod]: a structure containing the parameters of the spheroid
+    # some of the fields in [geod] are mildly duplicative of the information
+    # available in [rep], which can be exposed only through _get_spheroid_defn
+
+    # It's needed as an argument for computing great circle distances though,
+    # and for most applications, users will only have to deal with a small number
+    # of projection objects, and so we precompute it for each Projection
+    geod::geod_geodesic
```
c42f: I don't anticipate ever needing this for my use case, so from my point of view it seems a bit unfortunate to always allocate this.

c42f: Or should I say, always initialize it.  May be insignificant in practice I suppose.

src/proj_functions.jl
```diff
@@ -0,0 +1,211 @@
...
+    dest     - destination after moving for [distance] metres in [azimuth] direction.
+    azi      - forward azimuth (degrees) at destination [dest].
+
+""" ->
+geod_direct(position::Vector{Float64}, azimuth::Float64, distance::Float64, proj::Projection) = 
+    geod_direct!(copy(position), azimuth, distance, proj)
+
+@doc "Returns the destination by moving along the ellipsoid in the given projection" ->
+destination!(position::Vector{Float64}, azi::Float64, dist::Float64, proj::Projection) = geod_direct!(position, azi, dist, proj)[1]
+destination(position::Vector{Float64}, azi::Float64, dist::Float64, proj::Projection) = destination!(copy(position), azi, dist, proj)
```
c42f: This function name is pretty generic.  It's got a `Projection` as an argument, so all is not lost, but still.  If it wasn't so late, I'd try thinking of an alternative ;-)

c42f:
I haven't got through all this yet, it's quite a chunk of code.  I'll have to revisit it another day.

Too bad that reviewing code isn't as fun as writing code!

src/proj_functions.jl
yeesian: I was mostly using `turf-destination` as prior art for the name. (They assume WGS84, so they don't need a projection argument.) But no, it isn't too late to think of a different name.

src/proj_types.jl
yeesian: It is unfortunate/wasteful (but not by much) in my mental model of a single projection for an entire dataset. Two alternatives here, neither particularly desirable to me:

1. Make the user construct their own `Geodesic` object from the `Projection`, before performing the geodesic/distance calculations.
2. Construct (and then discard) the `Geodesic` object everytime the user performs a geodesic/distance calculation.

The first one incurs a great mental object cost (knowing about projections and ellipses/etc, and how to transform between them, taking care to convert between xy and lonlat) for the end-user.

The second one is going to be alot more expensive for people who do need elliptical distance calculations (e.g. OpenstreetMap/etc).

src/proj_functions.jl
yeesian: Hah yes

src/proj_functions.jl
yeesian: Yes, there is an explosion in the number of versions, but each does something different. The function-names are also clear and unambiguous in what they're representing.

What if you want to run code that's similar to
```julia
for i=1:1000000
   transform2!(src, dest, position)
end
```
(And I want/need to be able to do it for my use-cases [when working with OpenStreetMap/etc], when you're working with maps of vector data, and need to perform the projection for each point)

I cannot afford to be doing the check for whether its 2 or 3 dimensions within the loop itself. And there's no way for users to salvage the situation from a library function that insists on doing it, apart from having to roll out my own code.

The problem might go away in the long run, if/when fixedsizearrays comes to julia (and you can dispatch based on the type), but the need for it is immediate and urgent; do be sympathetic.

src/proj_capi.jl
```diff
@@ -0,0 +1,158 @@
...
+# functions, but without the pj prefix.
+
+immutable ProjUV
+    u::Cdouble
+    v::Cdouble
+end
+
+@doc "forward projection from Lat/Lon to X/Y (only supports 2 dimensions)" ->
+function _fwd!(lonlat::Vector{Cdouble}, proj_ptr::Ptr{Void})
+    xy = ccall((:pj_fwd, libproj), ProjUV, (ProjUV, Ptr{Void}), ProjUV(lonlat[1], lonlat[2]), proj_ptr)
```
yeesian: :ok_hand:

yeesian:
> I expect real life to intrude in the shape of a new baby within a couple of weeks or so!

Congratulations, I'm excited for you! :smile: Don't worry about the build-scripts/etc, since julia0.4 isn't even out yet

src/proj_functions.jl
yeesian: Reference: http://julia.readthedocs.org/en/latest/manual/style-guide/#handle-excess-argument-diversity-in-the-caller

yeesian:
Okay, regarding the discussion in https://github.com/FugroRoames/Proj4.jl/pull/4#discussion_r38784592, here's a small benchmark:

Setup (get `tech_square.osm` from `https://dl.dropboxusercontent.com/u/8297575/tech_square.osm`):
```julia
using Proj4, OpenStreetMap

wgs84 = Projection(Proj4.epsg[4326])
mass = Projection(Proj4.epsg[26986])

nodes,highways,buildings = getOSMData("tech_square.osm");
nodes = Vector{Float64}[Float64[lla.lon,lla.lat] for lla in collect(values(nodes))]
```

Before this PR:
```julia
julia> @time begin
              for i=1:1410
              transform!(wgs84,mass,nodes[i])
              end
              for i=1:1410
              transform!(mass,wgs84,nodes[i])
              end
              end
  0.002591 seconds (35.64 k allocations: 1.404 MB)
```

After this PR:
```julia
julia> @time begin
              for i=1:1410
              transform!(wgs84,mass,nodes[i])
              end
              for i=1:1410
              transform!(mass,wgs84,nodes[i])
              end
              end
  0.001608 seconds (4.62 k allocations: 248.406 KB)

julia> @time begin
              for i=1:1410
              transform2!(wgs84,mass,nodes[i])
              end
              for i=1:1410
              transform2!(mass,wgs84,nodes[i])
              end
              end
  0.001604 seconds (4.62 k allocations: 248.406 KB)
```

src/proj_functions.jl
yeesian: Okay, removed the multiplicity of transforms. Thanks for being persistent on this one. It took me a awhile to track down all the inefficiencies, which led to a much cleaner codebase in the end. See https://github.com/FugroRoames/Proj4.jl/pull/4#issuecomment-137867408 for a small benchmark

src/proj_functions.jl
c42f: Haha :-)  All I meant was the above was written at 1:30 in the morning and I was falling asleep - not my creative peak ;-)

src/proj_types.jl
c42f: Yes, none of the options seem entirely desirable.  How about the following alternative: define three low level types `Proj`, `ProjCtx` and `GeodCtx` (insert better names here) which are used in the low level functions instead of passing pointers.  You're already taking this route with the `geod_geodesic` type.  In general I feel a bit bad about passing pointers around in julia, even to the low level functions.

Having defined these, a `Projection` might become
```julia
type Projection
    ctx::ProjCtx
    proj::Proj
    geod::GeodCtx
end
```

If someone really wants control, they can use the low level interface and allocate their `Proj`, `ProjCtx`, `GeodCtx` independently.

Let's not do this in the current PR, it's already pretty huge.

src/proj_capi.jl
```diff
@@ -0,0 +1,163 @@
...
+    _errno() != 0 &&  error("forward projection error: $(_strerrno())")
```
c42f: Stylistic nitpick: I find this really hard to read.  The following "assertion style" equivalent works much better for me, since it can be read as a valid English sentence:

```julia
_errno() == 0 || error("forward projection error: $(_strerrno())")
```

src/proj_capi.jl
c42f: thanks

src/proj_functions.jl
c42f: In the turf example, it appears the usage would be namespaced as `turf.destination` - that's not hard to understand when reading code which uses turf.  For exported functions in julia it's less common to be explicitly writing namespaces, so IMO generic names should apply only to similarly generic functionality.

yeesian: Users can similarly do
```
import Proj4
Proj4.destination(...)
```
Tbh, I don't see what's ambiguous about destination, given the parameters and argument names, in the context of Proj4.

That said, let's switch to a better name if there's one.

src/proj_types.jl
yeesian: > Let's not do this in the current PR, it's already pretty huge.

You mean: not define
```julia
type Projection
    ctx::ProjCtx
    proj::Proj
    geod::GeodCtx
end
```
in this PR? I can see what you're saying, but (to my knowledge) it isn't common practice in julia wrapper libraries. I would write
```julia
type Projection
    ptr::Ptr{Void}
    geod::geod_geodesic
end
```
(e.g. https://github.com/JuliaGeo/LibGEOS.jl/blob/master/src/geos_types.jl) to make the correspondence to pointers obvious from the attribute name, but wouldn't go so far as to wrap it in additional types.

IMO, writing `proj.ptr` provides the same level of warning as writing `proj.rep.ctx`? And if a user is aware of the existence of the `proj.ptr` attribute, and intends to use it, that'll be undocumented behavior (and hence their responsibility)

src/proj_functions.jl
c42f: Of course, a user can do that, but if that's the recommended usage, should we really be exporting it?  (You'd think that `export` would define a module public interface, but that doesn't always seem to be the case in the developing best practice, for example `Base.Pkg` and `Base.Dates` don't export their public interfaces.  For some more discussion, including possibly adding a `public` keyword, see JuliaLang/julia#12647)

c42f: By the way, my objection here is not about ambiguity for the compiler - it's about readability for people consuming the API, and in particular reading code which uses the API.  If you encounter `destination(pos, 30, 1000, proj)` deep in a piece of code written by someone else, IMO it will be far from clear exactly what is being done.

To throw some alternative ideas out there

walk_geodesic
step_geodesic
stride_geodesic
project_geodesic
geodesic_step
geodesic_dest
geodesic_destination

Personally I find `walk_geodesic` brings a pretty clear picture to mind about what's going on.

c42f:
It's looking fairly solid, and I'm inclined to merge this soon rather than any later since it's basically blocking me from doing anything further (and probably you too).  I have more cleanup and decisions on the naming of API functions in mind, but those can be done later.

However, I tried to run the tests, and found that the `geod_` set of functions must be a relatively recent addition - they're not in my version of libproj as installed from the package manager (4.8.0).  It'd be good if the wrapper was graceful about not requiring a very recent proj version, given that the basic proj.4 functionality is extremely mature.  PyCall has an example of one way to do something similar - it does some switching on `pyversion`.

c42f:
It's great to have an example benchmark by the way.  It might be worth considering putting the test loops inside a function so that the optimizer has something to chew on and a chance to inline the calls to `transform!` if it sees fit.

yeesian:
> However, I tried to run the tests, and found that the geod_ set of functions must be a relatively recent addition - they're not in my version of libproj as installed from the package manager (4.8.0). It'd be good if the wrapper was graceful about not requiring a very recent proj version, given that the basic proj.4 functionality is extremely mature. PyCall has an example of one way to do something similar - it does some switching on pyversion.

Sure. They don't provide an easy way to query the major.minor version of Proj4 though, so we'll probably have to parse the string returned by `pj_get_release()`.

May I check what it returns for you? E.g.
```julia
julia> using Proj4
Proj4.
julia> Proj4.version() # calls pj_get_release()
"Rel. 4.9.1, 04 March 2015"
```

yeesian:
> However, I tried to run the tests, and found that the geod_ set of functions must be a relatively recent addition - they're not in my version of libproj as installed from the package manager (4.8.0). It'd be good if the wrapper was graceful about not requiring a very recent proj version, given that the basic proj.4 functionality is extremely mature.

Ah, that's our blocker, haha. No wonder you prefer the decoupling of the geodesic stuff from the Projections. (The issue is that even the initialization of the Projection (in its current state) already relies on a `geod_` function)

I'll see what I can do about it.

yeesian:
Give it a shot now? We'll probably move it into the build script when we eventually have one

yeesian:
> It's great to have an example benchmark by the way. It might be worth considering putting the test loops inside a function so that the optimizer has something to chew on and a chance to inline the calls to transform! if it sees fit.

For another PR/issue maybe? (:

c42f:
> > It's great to have an example benchmark by the way. It might be worth considering putting the test loops inside a function so that the optimizer has something to chew on and a chance to inline the calls to transform! if it sees fit.
>
> For another PR/issue maybe? (:

Sure.  I just meant: thanks for putting together the benchmark with tech_square.osm.

c42f:
The version installed with ubuntu 14.04:

```
julia> Proj4.libproj_version()
"Rel. 4.8.0, 6 March 2012"
```

yeesian:
Ah, the regex I wrote should work for your version too then. See if it runs (and passes) the relevant tests now?

c42f:
Tests pass now!  There's a couple of niggles left which I'll note on the associated lines.

src/proj_geodesic.jl
```diff
@@ -0,0 +1,152 @@
...
+
+    f = (a − b)/a = 1 − sqrt(1-es)
+    n = (a − b)/(a + b) = f /(2 − f)
+    es = (a^2 − b^2)/a^2 = f(2 − f)
+
+Reference: equations (1)-(3) of 
+    Algorithms for geodesics (arXiv:1109.4448v2 [physics.geo-ph] 28 Mar 2012)
+""" ->
+function _geod_geodesic(a::Cdouble, f::Cdouble)
+    g_ptr = pointer_from_objref(geod_geodesic())
```
c42f: This seems dangerous, and will likely cause segfaults later.  From the docs:

> pointer_from_objref(object_instance)
>
>  Get the memory address of a Julia object as a Ptr. The existence of the resulting Ptr will not protect the object from garbage collection, so you must ensure that the object remains referenced for the whole time that the Ptr will be used.

The critical thing being that we're not holding a julia reference to the `geod_geodesic` instance (rather, we're returning a bare pointer) so garbage collection will eventually collect the instance and our pointer will be left dangling.

src/proj_types.jl
```diff
@@ -0,0 +1,46 @@
...
+Cartographic projection type
+""" ->
+type Projection
+    #ctx::Context   # Projection context object
+    rep::Ptr{Void} # Pointer to internal projPJ struct
+    
+    # [geod]: a structure containing the parameters of the spheroid
+    # some of the fields in [geod] are mildly duplicative of the information
+    # available in [rep], which can be exposed only through _get_spheroid_defn
+    geod::Ptr{Void}
```
c42f: I suggest simply adding a dummy geod_geodesic type when the version of the underlying libproj isn't enough:

```julia
type geod_geodesic
    # underlying libproj not recent enough to support geod_geodesic
end
```

src/proj_functions.jl
yeesian: oops, i forgot about this. Shall we go with `geodesic_dest`, as one of the nicer-sounding nouns?

src/proj_types.jl
yeesian: :ok_hand: thanks for the catch! That was really careless of me ><

src/proj_functions.jl
yeesian: or `geod_destination` (and `geod_distance`) for consistency

c42f:
Ah cool you fixed the REQUIRE, I had to tweak that locally.

Personally I think the only really good thing about the `geod_*` names is that they're consistent with the underlying library.  `geod_destination` is definitely better than `destination` though.

c42f:
Having said that, I think I'll merge this now, and do a full review of the API naming consistency once it's clearer what exactly is going to be included.

c42f:
Oh, I like that you renamed `ellps_distance` for consistency.  I didn't notice that immediately.

c42f:
Ok, I rebased and merged.  Thanks for all the hard work.

yeesian:
:+1: Thanks for your code reviews! the resulting code has significantly improved for it

c42f:
Cheers :)

c42f:
Oh, BTW I did rebase your changes before merging to make the history simpler.  It might be worth it if you reset your master before doing anything more, just to keep the history tidy.  I hope that's not confusing, it's always hard to know how much attention to pay to keeping the history graph readable.

yeesian:
I'm afraid it is confusing (I could always nuke my repo, and re-fork); if it's not too much trouble, can you do a set of instructions for me to follow (after which I'll probably have a good sense of what's happening)?

c42f:
Ack sorry about that, now that I've been using if for quite a while I sometimes forget how confusing git is.

All you should need to do is reset your master branch to FugroRoames/master:

```bash
git checkout master
# Now, ensure you don't have any local changes on master or in the working copy!
# The following resets the current branch (master) to point to FugroRoames/master,
# assuming your git remote name for the FugroRoames/Proj4 is actually FugroRoames.
git reset --hard FugroRoames/master
```

After that, branch any new changes off the top of your master.

PR #6, opened 2015-09-24
install libproj for OSX (homebrew) & Linux (apt-get)

sebastiang:
more to do, but this gets the ball rolling for the two platforms I have easy access to.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,0 +1,19 @@
+using BinDeps
+
+@BinDeps.setup
+
+libproj = library_dependency("proj", aliases = ["libproj"])
+
+@linux_only begin
+    provides(AptGet, "libproj0", libproj, os = :Linux)
+end
+
+@osx_only begin
+    if Pkg.installed("Homebrew") === nothing
+        error("Homebrew package not installed, please run Pkg.add(\"Homebrew\")")
+    end
+    using Homebrew
+    provides(Homebrew.HB, "proj", libproj, os = :Darwin)
+end
+
+@BinDeps.install
--- a/src/Proj4.jl
+++ b/src/Proj4.jl
@@ -1,5 +1,8 @@
 module Proj4
 
+using BinDeps
+@BinDeps.load_dependencies
+
 const libproj = "libproj"
 
 export Projection, # proj_types.jl
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 *.jl.cov
 *.jl.mem
+deps/deps.jl
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,1 +1,3 @@
 julia 0.4-
+BinDeps
+@osx Homebrew


sebastiang:
hold up; I pushed a bad commit. I'll fix this shortly.

c42f:
Awesome, thanks for having a go at this :)

deps/build.jl
```diff
@@ -0,0 +1,19 @@
+using BinDeps
+
+@BinDeps.setup
+
+libproj = library_dependency("proj", aliases = ["libproj"])
+
+@linux_only begin
+    provides(AptGet, "libproj-dev", libproj, os = :Linux)
```
c42f: Maybe we should be installing only the binary dependencies - that would be "libproj0" rather than "libproj-dev"

deps/deps.jl
```diff
@@ -0,0 +1,11 @@
+# This is an auto-generated file; do not edit
+
+# Pre-hooks
+
+# Macro to load a library
+macro checked_lib(libname, path)
```
c42f: Where does this get used?  Is it somewhere automagically required by BinDeps, or should we be doing an extra include inside "Proj4.jl" ?

sebastiang:
see what this does for you. seems to pass tests. Though `libproj0` in Ubuntu 14.04 is from 2012!

c42f:
It's true that `libproj0` is old, but installing `libproj-dev` doesn't fix this: the `-dev` version only has headers and associated development-related stuff like static libraries.  The actual shared object that Proj4.jl links against is from the `libproj0` package.

src/Proj4.jl
```diff
@@ -1,5 +1,8 @@
 module Proj4
 
+using BinDeps
+@BinDeps.load_dependencies
+
 const libproj = "libproj"
```
c42f: Should we now remove the `const libproj = "libproj"` from here, as `BinDeps` will define it for us in `load_dependencies`?

sebastiang:
indeed. just surprised that when you grab `proj` with Homebrew on the Mac you get something from 2 weeks ago, while on Ubuntu you get it from 3 years ago. The difference is enough to not get the geodesic support on Linux. I'm not wise in the ways of package management, but it might be enough to cause grief with users.

src/Proj4.jl
sebastiang: When I am better rested, or the packages are updated for the last bits of v0.4 deprecations, I will understand just what the options are for `load_dependencies`, as indeed I believe it will define a handle to the library. In my mind this was related to doing precompilation effectively which I didn't want to tangle this up with.

c42f:
Agreed that ubuntu is rather out of date with this.  An alternative is to build the binary dependency ourselves using the `Sources` provider.  Shouldn't be too hard I guess - proj.4 itself seems to provide a sane looking cmake setup.

src/Proj4.jl
c42f: Yes, I don't yet know the details of how `BinDeps` relates to precompilation.  The current setup with the string constant for `libproj` seems clearly ok for precompilation, whereas caching a pointer to a `dlopen()`'d libproj would clearly be bad ;-)  I don't know if BinDeps does something sane here yet or not...

c42f:
BTW, I just pushed some changes to make sure the unit tests actually run (they were broken when porting to FactCheck.jl, in a way that gave complete false confidence: running the tests didn't actually test any of the `@fact` assertions!)  I'll be cleaning up the resulting errors...

sebastiang:
OK, that will be great. The examples I tried didn't like arrays of `Int64` and I didn't understand why. Will look into it today.

sebastiang:
Chris, I'm inclined to leave this as is, as it's good enough for my purposes and I need to move on to other things! Building from source sounds like a good idea in time, and I'm sure someone will want Windows support before too long, but those are beyond my needs.

c42f:
Looks pretty good other than the mismatch between `load_dependencies` and the const libproj string which seems a bit odd to me.  I'll test it out locally and see whether something else makes sense.

I've pushed some bug fixes which should make the minimal example work again.

sebastiang:
My gut is the right time to fix that mismatch is when you build Windows support. Linux and OS X are close enough that usually the “libXXX” protocol works. On Windows it tends to be “XXX.dll”, or something weirder, and then it’ll be nice to get the name from BinDeps.

c42f:
The point about `load_dependencies` is that since we're not using the `library_dependency` in the main code yet, calling `load_dependencies` in the main `Proj4.jl` doesn't actually do anything at this stage, other than force Proj4 to load BinDeps as a (rather large) runtime dependency.

With that in mind, I did a small tweak to comment that part out for now before merging.  I understand this is orthogonal to the way that `Pkg.build()` is called so there should be no change in functionality.

c42f:
Thanks a lot, rebased + merged.  I'll tackle the windows build later unless some other random hero steps in :)

yeesian:
Late to the party here, but am glad to see things moving! (:

PR #7, opened 2015-09-26
Support for Tuples

yeesian:


c42f:
Thanks.  Pardon the travis failures - I've just enabled it for this repo and am figuring out how to set it up.

c42f:
Hmm, supporting tuples as well as vectors and matrices makes me start to worry about the combinatoric explosion of different argument types a user may want to pass to `transform()` and friends.  For example, personally I'm more likely to want to pass a `Vector2` or `Vector3` type from ImmutableArrays.jl to these functions (and what about `Vector{Vector3}` for a vectorized version?)

Obviously we can't implement special versions for everything people might want to pass in.  Maybe there's a way to share an implementation between all kinds of short-vector-like things?  One sensible option might be a traits based approach where we require `lon()`, `lat()`, `x()`, `y()`, `height()`, etc for any short-vector-like thing.

Is there a specific use case for wanting tuples as a blessed type in the interface?  AFAIK a vector of tuples isn't a particularly efficient construct at the current time (though from what I'm seeing, this will be fixed during 0.5).

yeesian:
This was originally begun by [some work on OpenStreetMap](https://github.com/yeesian/OpenStreetMapParser.jl/commit/641ba1fbead49ace57e45a839703820b04266a46#diff-8bd2881ef7e8a7600c2c48d9539d2e7eR72) where the nodes are `Tuple{Float64, Float64}`. I have since removed OSM's dependency on Proj4, so I'll be okay with closing this PR.

Your point about the amount of work it'll take to support multiple different formats is well-taken. The notions of traits is something I had thought about (e.g. in [GeoInterface](https://github.com/JuliaGeo/GeoInterface.jl)), but I haven't been following the development in Array-land very carefully, and didn't want to invest too much time into pursuing it.

For now, I'm not ready to begin the work of supporting more and more formats (Vec2, Vec3, etc) to the point it will make sense to use some meta-programming to auto-generate the code, so I'll be okay with closing this PR.

c42f:
Right, I've rebased+merged the first two commits here, and left the tuple changes for later.  I've pushed them to tuple-support for now so they don't get lost, but I'd prefer to work out a generic interface rather than implementing things in Proj4 for a grab bag of types.  (When I mentioned ImmutableArrays above, I don't mean Proj4.jl should support that directly - rather it should supply an interface to make it easy for users to use their own types.)

The issue of the most appropriate interface for geospatial points is an interesting one, and not something that Proj4.jl can decide on in isolation from the rest of the geospatial libraries.  I like the general idea of Proj4.jl depending on GeoInterface when it's ready.

PR #10, opened 2016-02-29
Building libproj from source on Linux

awbsmith:
###### The Good:

- It'll build Proj4 from source allowing for newer Proj4 versions than apt-get

###### The Bad:

- No minimum version for the libproj dependency (if you already have an old libproj version it won't build a newer one)
- I've only done it for Linux
- Slower load time and pre-compiling whinges.  
- There's got to be a way to get the library name and path from the dependency entry

###### The Other

- had to rename the library name variable to not conflict with the dependency entry

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -10,13 +10,12 @@ libproj = library_dependency("libproj", aliases = ["libproj"])
 @linux_only begin
 
 	
-	
 	# AptGet's LibProj is old
 	#provides(AptGet, "libproj0", libproj, os = :Linux)  
 
 	# so download a newer version
 	libsrc = "http://download.osgeo.org/proj"
-	libproj_ver = "4.9.1" # from source http://download.osgeo.org/proj/
+	libproj_ver = "4.9.2" # from source http://download.osgeo.org/proj/
 	libproj_name = "proj-$(libproj_ver).tar.gz"
 
 	# and build it from source
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -24,18 +24,17 @@ const libproj_ver = "4.9.1"
 			GetSources(libproj)
 			CreateDirectory(builddir)
 			@build_steps begin
-				CreateDirectory(prefix)
-				println(srcdir)
-				cd(srcdir)
-				FileRule(joinpath(prefix,"lib","libproj.so"), @build_steps begin
-					#`./configure --prefix="$prefix"`
-					`./configure`   # installs to usr/local/lib by default
+				ChangeDirectory(builddir)
+				FileRule(joinpath("/usr", "local", "lib","libproj.so"), @build_steps begin
+					#`echo Configuring: $(srcdir)/configure --prefix=$(prefix)`
+					#`$(srcdir)/configure --prefix=$(prefix)`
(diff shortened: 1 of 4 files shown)


awbsmith:
Hmm, the Ubuntu [Universe](http://www.ubuntuupdates.org/package/core/wily/universe/base/libproj-dev) ppa has libproj 4.9.1 anyway.  There might have been an easier way...

deps/build.jl
```diff
@@ -2,10 +2,45 @@ using BinDeps
...
 @linux_only begin
-    provides(AptGet, "libproj0", libproj, os = :Linux)
+
+	
+	# AptGet's LibProj is old
+	#provides(AptGet, "libproj0", libproj, os = :Linux)  
+
+	# so download a newer version
+	libsrc = "http://download.osgeo.org/proj"
+	libproj_ver = "4.9.2" # from source http://download.osgeo.org/proj/
```
nigels-com: It's nice to be using a specific version, not just whatever the latest release is.

nigels-com:
Does it work on RedHat, CentOS or Fedora?

BuzzVII:
the big question is does it work on the Amazon linux ami. we need all these
to build and package in prod.

nigels-com:
It's more likely to work on Amazon Linux (or anywhere else) if it isn't relying on some specific version installed system-wide, right?

awbsmith:
Lets find out!

deps/build.jl
```diff
@@ -2,10 +2,45 @@ using BinDeps
...
+	libproj_name = "proj-$(libproj_ver).tar.gz"
+
+	# and build it from source
+	prefix   = joinpath(BinDeps.depsdir(libproj), "usr")
+	srcdir   = joinpath(BinDeps.depsdir(libproj), "src", "proj-$(libproj_ver)")
+	builddir = joinpath(BinDeps.depsdir(libproj), "builds", "libproj-$(libproj_ver)_version")
+	libdir   = joinpath(prefix, "lib")
+
+	provides(Sources, URI(joinpath(libsrc, libproj_name)), libproj, os = :Linux, installed_libpath=libdir)
```
c42f: We should now be able to get away with `:Linux` -> `:Unix`, and `@unix_only`

deps/build.jl
```diff
@@ -2,10 +2,45 @@ using BinDeps
...
+
+	provides(BuildProcess,
+		(@build_steps begin
+			GetSources(libproj)
+			CreateDirectory(builddir)
+			@build_steps begin
+				ChangeDirectory(builddir)
+				FileRule(joinpath(libdir, "libproj.so"), 
+				@build_steps begin
```
c42f: Let's indent as

```julia
FileRule(joinpath(libdir, "libproj.so"),
         @build_steps begin
            ChangeDirectory(srcdir)
            `./configure --prefix=$(prefix)`
            `make`
            `make install` 
         end)
```

c42f:
@awbsmith argh.  Apparently I managed to push this all to master, along with my cleanups when I meant to push my extra changes to a branch.

Never mind, I think it's all good...  let me know if things are not as they seem.

PR #11, opened 2016-09-22
Removed Julia 0.5 deprications

mauro3:


src/proj_capi.jl
```diff
@@ -69,7 +69,7 @@ end
 
 "Get human readable error string from proj.4 error code"
 function _strerrno(code::Cint)
-    bytestring(ccall((:pj_strerrno, libproj), Cstring, (Cint,), code))
+    @compat unsafe_string(ccall((:pj_strerrno, libproj), Cstring, (Cint,), code))
```
c42f: This is not a syntax change, so `@compat` should not be needed here - just `using Compat` should do the trick (From Compat's README it looks like we'll need at least Compat v0.8.0 in the REQUIRE to get `unsafe_string`).

mauro3: Good point, fixed.

c42f: Perfect, thanks.

.travis.yml
```diff
@@ -3,8 +3,9 @@ os:
   - linux
   - osx
 julia:
-#  - release # 0.3 not supported yet
-  - nightly
+  - 0.4
+  - 0.5
+# - nightly
```
tkelman: instead of commenting this out, make it an allowed failure so it'll run but won't make the status red?

tkelman: https://docs.travis-ci.com/user/customizing-the-build#Rows-that-are-Allowed-to-Fail

mauro3: Ok, for the travis-challenged like myself:
```
...
julia:
  - 0.4
  - 0.5
  - nightly
matrix:
  allow_failures:
    - julia: nightly
...
```

PR #12, opened 2017-01-19
More 0.5 deprecations removed

mauro3:
Last time (#11), I forgot the ones in the bulid.jl.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,4 +1,4 @@
-using BinDeps
+using BinDeps, Compat
 
 @BinDeps.setup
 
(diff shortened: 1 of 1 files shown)


c42f:
Travis now fails on 0.4 (`@static` isn't present, etc).  `using Compat` might fix that?

mauro3:
Fixed.

deps/build.jl
```diff
@@ -51,7 +51,7 @@ provides(BuildProcess,
 )
 
 
-@osx_only begin
+@compat @static if is_apple()
```
c42f: I'm being picky here, but the explicit `@compat` shouldn't be required here - these are not syntax incompatibilities, but API incompatibilities.  So `Compat` will define both `unsafe_string` and `@static` for older versions.

visr: And being even more picky, since this is just a build script there's not really a performance benefit to have it evaluated at parse time, so the `@static` could be left out altogether.

mauro3: Thanks for calling out my `Compat` ignorance! Updated.

c42f:
Thanks for the fixes!

PR #17, opened 2018-03-25
use BinaryProvider

### LINKED ISSUE #8: Install on windows + CI setup
Getting Proj4.jl to work on windows mainly involves building the C library.  This is pretty easy as a one-off, but then it'll need to be distributed from somewhere, etc, which is a bit of a pain.  OSGeo4W comes to mind.  Just grabbing it from their release area might possibly be ok, or maybe it should be done via their installer.  Latest release is here:

http://download.osgeo.org/osgeo4w/x86_64/release/proj/

visr:
I've tried out BinaryBuilder.jl at https://github.com/visr/PROJBuilder
It seems to work well, so now I wanted to try using it here, and fix #8.

Note that this also updates PROJ from (at least) 4.9.0 to (always) 4.9.3. The `epsg` and `esri` files are reprocessed to match this. With this I'm getting `errorFraction` in the tests of 3.88%. Are we sure this test is valid though, that if pulling it through 3 times doesn't result in the same string, it is by definition an error? Since now we know that both the files and binaries are from the same PROJ version.


visr:
Regarding the 3.88% error rate, they are all from `esri`, none from `epsg`.

CI is green, AppVeyor too.
What do you think about this?
Perhaps it would be good to move PROJBuilder to JuliaGeo before merging this.

yeesian:
I'm on board with both moving PROJBuilder to JuliaGeo and merging this.

yeesian:
Thanks!

PR #18, opened 2018-04-28
use latest build and BinaryProvider

visr:
The latest BinaryProvider allows updating already installed tarballs, and writes an updated `build.jl` such that I can use it as a dependency for `GDALBuilder`.
The build configuration itself did not change however. But still new hashes, a later BinaryBuilder was used to create them.

https://github.com/JuliaGeo/PROJBuilder/compare/v4.9.3-0...v4.9.3-1


yeesian:
LGTM

visr:
Didn't realise I could just use the build script from the builder releases page... Using that now. Should be done now.

deps/build.jl
```diff
@@ -1,42 +1,53 @@
...
+        install(url, tarball_hash; prefix=prefix, force=true, verbose=verbose)
     end
-
-    # Finally, write out a deps.jl file
-    write_deps_file(joinpath(@__DIR__, "deps.jl"), products)
+elseif unsatisfied
+    # If we don't have a BinaryProvider-compatible .tar.gz to download, complain.
+    # Alternatively, you could attempt to install from a separate provider,
+    # build from source or something even more ambitious here.
+    error("Your platform $(triplet(platform_key())) is not supported by this package!")
```
yeesian: maybe the comment should go into the error message?

deps/build.jl
```diff
@@ -1,42 +1,53 @@
-using BinaryProvider
+using BinaryProvider # requires BinaryProvider 0.3.0 or later
```
yeesian: the comment is already reflected in REQUIRE

visr:
Regarding your comments @yeesian, the `build.jl` is now an unaltered version of https://github.com/JuliaGeo/PROJBuilder/releases/download/v4.9.3-2/build.jl, which is created  by BinaryBuilder. I think it would be best to keep them 1-1, since these files are also used when it is used as a dependency for GDALBuilder.

So then it would be best to raise these issues directly with BinaryBuilder.
Same goes for https://github.com/JuliaGeo/LibGEOS.jl/pull/49

yeesian:
oh okay got it. anyhow, it's not a blocker for this PR. thanks!

PR #19, opened 2018-07-06
upgrade to julia 0.7

visr:
@c42f I increased the tolerance here quite a bit: 4e0897acb49635578940463f3e8925f6f3442a89, wanted you to be aware.
I simply lowered it one order of magnitude at a time until all 4 of these tests passed. Considering these are very large numbers, I guess this is fine. Using the default arguments (no abs, only rel err), also passes.

We can perhaps increase it again if we support PROJ 5, from [gdalbarn.com](https://gdalbarn.com/):
> PROJ 5.0+ no longer requires datum transformation pivots through WGS84, which can introduce errors of up to 2m, but the rest of the tools do not take advantage of it.


yeesian:
LGTM

PR #23, opened 2019-05-18
switch to a PROJ 6.1.0 build

visr:
PROJ has changed quite a bit in the past year, largely as a result of plans laid out in the GDAL barn raising ([gdalbarn.com](https://gdalbarn.com/)). For the build this means a dependency on SQLite, since the CSV EPSG file is no more, and the data now lives in a SQLite database named `proj.db`.

I started this build mainly since in GDAL 3.0 PROJ 6+ is needed, and I want a GDAL 3 build. I see from https://proj4.org/news.html that this also removes `projects.h` and deprecates `proj_api.h` (a new API is introduced).

This commit only updates the build. Rewrapping the APIs with Clang.jl can be done separately. Also `generate_projection_codes.jl` is not rerun, since it no longer works, since it should make use of `proj.db`. Updating that can also be done separately.

This also removes REQUIRE and adds Project.toml. I raised the minimum julia version to 1.0, since 0.7 is past end of life, or are there any objections to this?


yeesian:
LGTM.

> Rewrapping the APIs with Clang.jl can be done separately. Also generate_projection_codes.jl is not rerun, since it no longer works, since it should make use of proj.db. Updating that can also be done separately.

Can you file these as issues if this gets merged?

c42f:
Nice! I've been out of this for a while, but it's great to see some of the big problems which *seemed* permanently baked into the proj.4 API are being resolved.

By the way, I just saw the release notes for version 5.0, with their decision to change names: https://proj4.org/news.html#versioning-and-naming

With this in mind it probably makes sense to also change the name of Proj4.jl in the future.

visr:
> With this in mind it probably makes sense to also change the name of Proj4.jl in the future.

Yeah indeed I was thinking the same thing. Let's rename it to `PROJ.jl`. Now that we switched to the new registry this should be easier, as it can keep the same UUID. Though didn't see any renames yet in General.

PR #27, opened 2019-05-31
use Clang to wrap PROJ 6.1.0

visr:
I plan to finish https://github.com/JuliaGeo/GDAL.jl/pull/66 first, but since that is now stuck on PROJ_LIB related issues, it's convenient to have the new API at hand. Lessons learned there will probably have to be applied here too.

ref https://discourse.julialang.org/t/setting-env-for-a-shared-library-in-init/24724/


c42f:
Very cool :tada:

visr:
Ha, with the new rewriter it becomes much more fun to rewrite wrapped functions. In d1d4bb5bbc624442abf830362f4c02d22cee3081 I made all `ctx` arguments keywords, also updating the order of parameters in the docstrings (though that means they no longer match C).

Here is the [doxygen.xml.gz](https://github.com/JuliaGeo/Proj4.jl/files/3499004/doxygen.xml.gz) for if anyone wants to run it.

We could take it further and do this for all essentially optional parameters. That would make some functions much easier to use, for instance `proj_coordoperation_get_param`:
https://github.com/JuliaGeo/Proj4.jl/blob/d1d4bb5bbc624442abf830362f4c02d22cee3081/src/proj_c.jl#L1345-L1382

Some things left to do:
- [ ] add tests to see if this new API is actually working
- [ ] `__init__`: [create a context](https://proj.org/development/reference/functions.html#c.proj_context_create)
- [ ] `__init__` point it to proj.db, similarly to [GDAL](https://github.com/JuliaGeo/GDAL.jl/blob/b75ac1c3a351ac934e35827289a0d522a1fa7005/src/GDAL.jl#L71-L73) but using PROJ's `proj_context_set_search_paths`.
- [ ] use [error reporting](https://proj.org/development/reference/functions.html#error-reporting). I think we should always check automatically for all ccall that return `Ptr{PJ}`. If it is an error, we get the error string and throw a `PROJError`

visr:
Ok I still added some error handling support. If PROJ throws an error that is turned into a Julia exception. Note that I am currently not checking each `Ptr{PJ}` that is returned from `ccall`, since I'm not sure if it is needed, and could not find a testcase that would not already have PROJ throw an error here.

Furthermor I still attempted to make using this interface a bit more pleasant, by turning some arguments into keyword arguments defaulting to C_NULL, when the docs stated this was accepted as well.

Test coverage is not yet great, but what we did test works well, and actually I think this is large enough a PR as is. @yeesian do you think this is ok to merge? No rush.

yeesian:
This is incredible work, thanks!

Yeah I think it is god enough to be merged; we can increase coverage after the fact (like with LibGEOS)

test/proj6api.jl
```diff
@@ -0,0 +1,69 @@
...
+            Acc:$(info.accuracy)
+        """)
+        Proj4.proj_destroy(operation)
+    end
+    Proj4.proj_list_destroy(results)
+
+    Proj4.proj_operation_factory_context_destroy(factory)
+    Proj4.proj_destroy(src)
+    Proj4.proj_destroy(tgt)
+    # proj_context_destroy(c)
```
yeesian: is this line meant to be removed?

test/proj6api.jl
```diff
@@ -0,0 +1,69 @@
...
+end
+
+@testset "Database" begin
+    crs = Proj4.proj_create_from_database("EPSG", "4326", Proj4.PJ_CATEGORY_CRS, false)
+    Proj4.proj_errno(crs)
+    Proj4.proj_get_id_code(crs, 0)
+    # The following is wrong, but unfortunately segfaults. How to fix other than adding
+    # a null check before the ccall in this and other functions? PROJ documentation
+    # clearly states: must not be NULL
+    # Proj4.proj_get_id_code(C_NULL, 0)
```
yeesian: we should file this as an issue (to be resolved outside of this PR [after it gets merged])

src/error.jl
```diff
@@ -0,0 +1,26 @@
+"Exception type based on PROJ error handling"
+struct PROJError <: Exception
+    msg::String
+    # reset PROJ's error stack on construction
```
yeesian: nit: consider moving this into the line right above `proj_errno_reset(C_NULL)`, so that it is more clearly associated with `function PROJError(msg)` than with `msg::String`

src/Proj4.jl
```diff
@@ -45,4 +43,37 @@ include("proj_functions.jl") # user-facing proj functions
...
+        push!(strings, unsafe_string(cstring))
+        i += 1
+        cstring = unsafe_load(ptr, i)
+    end
+    strings
+end
+
+const PROJ_LIB = Ref{String}()
+
+# Module initialization function
```
yeesian: is there a policy on not having docstrings for `__init__`? Otherwise, consider promoting the comment into a docstring.

gen/wrap_proj.jl
```diff
@@ -0,0 +1,154 @@
...
+        arg = fargs2[i]
+        fargs2[i] = :($arg = C_NULL)
+        # in optpos is does not have to be at i anymore if it already was moved
+        argoptpos = findfirst(==(i), argpos)
+        splice!(argpos, argoptpos)
+        push!(argpos, i)  # add it to the end
+    end
+end
+
+"Rewrite expressions in the ways listed at the top of this file."
```
yeesian: nit: Rewrite expressions using the transformations listed at the top of this file.

visr:
> it is god enough

High praise indeed! Ok merging :)

visr:
Thanks for the comments Yeesian, much appreciated. Addressed them in the latest commit.

PR #29, opened 2019-08-23
Add tests for new API

yeesian:


visr:
Nice work, thanks. I'll merge it onto the wrap6 branch, we can always still tweak things a bit there.

PR #33, opened 2019-12-05
use PROJ_jll for binary

visr:
I added the PROJBuilder build recipe to Yggdrasil, this led to PROJ_jll. Trying it out here.

https://github.com/JuliaGeo/PROJBuilder
https://github.com/JuliaPackaging/Yggdrasil/tree/master/P/PROJ
https://julialang.org/blog/2019/11/artifacts

**TODO**
- [x] C++ string ABI issue: https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/539
- [x] Create Artifact for the proj-datumgrid that were included in the build before, but now should be in a separate lazy artifact download. For an example see the blog above, and [this Artifact.toml](https://github.com/r3tex/ObjectDetector.jl/blob/master/Artifacts.toml) and [this script](https://github.com/r3tex/ObjectDetector.jl/blob/master/dev/generate_artifacts.jl)
- [ ] ~~Integrate the artifacts with the rest of the package and tests it~~ (Created an Artifacts.toml with all latest datumgrid releases that is downloaded on demand. Probably best to go the [PROJ 7](https://github.com/OSGeo/PROJ/releases/tag/7.0.0) route for further integration.)


c42f:
Very cool. Note that since proj is (IIRC?) fairly minimal on the dependencies it should be possible to support older julia versions at the same time (if you had the need) once https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/533 has propagated into the generated jll's.

visr:
PROJ now depends on SQLite, and the next release will probably rely on libtiff as well.

It should probably be doable to support older julia releases at the same time, given some work. I don't personally have the need though, and unless somebody is willing to make it happen, I'd be happy to drop older versions, given my limited time. They will still work as is, and we can always still make bugfix releases for the older versions as well.

I have no rush to merge this PR, for me it is mostly a stepping stone to finally getting a GDAL build with essential drivers that is not plagued by symlink issues for Windows users without administrator rights. Having that will also allow finally merging https://github.com/yeesian/ArchGDAL.jl/pull/76, which is stuck behind a broken GDAL GEOS in Linux, another build issue.

Note that PROJ_jll does not include the datumgrid artifact that was included by default before. I still want to make this a lazy artifact, that will download it when needed. Though by the time PROJ 7 hits, these large grid shift files can also be accessed remotely, if [PROJ RFC 4](https://github.com/rouault/PROJ/blob/rfc4_remote_and_geotiff_grid/docs/source/community/rfc/rfc-4.rst) lands.

visr:
I updated the artifacts with the latest proj-datumgrid packages that were released today.
PROJ 7 was also released today, see the [release notes](https://github.com/OSGeo/PROJ/releases/tag/7.0.0). Since it provides a new way to interact with the gridded models, it's probably best to refrain from working too much on the old system now.

PROJ 7 still supports the old `proj_api.h` API, but will be the last to do so.

I think it's best to just merge this as is. Anyone still wants to review?

src/Proj4.jl
```diff
@@ -30,12 +27,7 @@ end
 const version = _version()
 
 # Detect underlying libproj support for geodesic calculations
-const has_geodesic_support = version >= v"4.9.0"
-
-if has_geodesic_support
-    export geod_direct, geod_inverse, geod_destination, geod_distance
-    include("proj_geodesic.jl") # low-level C-facing functions (corresponding to src/geodesic.h)
-end
+const has_geodesic_support = true
```
felixcremer: Do we still need this?
If this is not going to change the value, we could delete this variable overall and also delete the references in the tests and in proj_functions.jl
If this is a wanted change, I could prepare a pull request.

visr: Thanks for having a look!

Correct, we don't need this internally anymore. At the same time, I don't know if people are using this, and it is fairly harmless to keep around for now.

I guess this can go once we move the package over from the old API to the new one, as in https://github.com/JuliaGeo/Proj4.jl/issues/34#issuecomment-573400435.

PR #35, opened 2020-01-16
Implement convenience constructors for common projections

asinghvi17:
Implemented so far:
`WGS84, Mercator, TransverseMercator, UTM, AlbersEqualArea, Robinson, WinkelTripel`

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,3 +19,14 @@ transform(wgs84, utm56, [150 -27 0])
 
 API documentation for the underlying C API may be found here:
 https://proj.org/development/reference/index.html
+
+Proj4.jl also provides some convenience functions to generate common projections; use the inline help mode in the REPL to find out more.  Currently, we supply the following constructors:
+```julia
+WGS84
+Mercator
+TransverseMercator
+UTM
+AlbersEqualArea
+Robinson
+WinkelTripel
+```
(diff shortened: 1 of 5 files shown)


visr:
Thanks! My main concern is that this wraps the old PROJ 4 API, which will be deprecated. Best to build on the new one.

asinghvi17:
Oh, `Projection` is being deprecated too?  Thought it was just the C API...will see what I can do to change it.

visr:
Yeah I meant the C API. But Projection is now built on top of the C API that will be deprecated. Though perhaps we can switch out the implementation of Projection instead.

asinghvi17:
Yeah I'd rather do that, since it's a pretty convenient thing to have - should be simple enough to do with the new API.

asinghvi17:
https://proj.org/development/reference/functions.html?highlight=conversion#coordinate-transformation

This describes the C API for conversion between CRS.

asinghvi17:
https://proj.org/development/reference/functions.html?highlight=conversion#coordinate-transformation

This describes the C API for conversion between CRS.

asinghvi17:
Issues that still remain with this PR are:
- [ ] No real system to specify the datum / ellipsoid
- [ ] Moving to the Proj 6 API

I might just close this, and create a separate PR which transitions to the Proj 6 API first.

visr:
I'll just close this, since it is targeting an API that is about to be removed.

I'm all for making it more convenient to construct projections, though I also worry about the maintenance burden of going down the route proposed here. E.g. which ones are important enough to get their own function? There are so many. A good start would be better documentation, and links to PROJ documentation like https://proj.org/operations/projections/index.html.

PR #36, opened 2020-01-27
Pass pointers or vectors to `pj_transform`

asinghvi17:


visr:
Why would you want this change, and do you still want it? I'd assume it was to be able to convert a set of points at once, but if I see https://github.com/JuliaGeo/Proj4.jl/blob/63aa9adc1459aa48454813d8da86a85ac8101c68/test/runtests.jl#L31-L52 it looks like that is already supported?

This PR is currently failing since the module specifier `Proj4.` was added in two places.

c42f:
Hmm interesting.  It seems that all the old API proj wrappers are incorrect because they should be passing the input arrays directly to `ccall` rather than passing pointers (this is likely my mistake from when I originally wrote a lot of the wrappers by hand in julia 0.4 or so!). It seems basically everything in `proj_capi.jl` is making this mistake so all arrays may not be rooted correctly during the `ccall`.  Mind you, this seems somewhat unlikely to result in an actual crash but it could in theory.

Maybe it's a sign that the old API should just be deleted anyway? The new `proj_` APIs are just so much better from a technical geodesy perspective. Though I guess such a change would motivate finally creating a new library `Proj.jl` and moving future development there!

visr:
> The new proj_ APIs are just so much better from a technical geodesy perspective. Though I guess such a change would motivate finally creating a new library Proj.jl and moving future development there!

The new API is of course already available here in `proj_common.jl` and `proj_c.jl`. But PROJ 7 is out now, and will be the last one to include the deprecated API, so it is definitely time that we move as well.

This is the [quick start](https://proj.org/development/quickstart.html) using the new API:
```julia
pj = Proj4.proj_create_crs_to_crs(
    "EPSG:4326",  # source
    "+proj=utm +zone=32 +datum=WGS84",  # target, also EPSG:32632
)
pj_for_gis = Proj4.proj_normalize_for_visualization(pj)
Proj4.proj_destroy(pj)
pj = pj_for_gis

a = Proj4.proj_coord(12, 55, 0, 0)
b = Proj4.proj_trans(pj, Proj4.PJ_FWD, a)

b = Proj4.proj_trans(pj, Proj4.PJ_INV, b)
Proj4.proj_destroy(pj)
```

It would be nice to add some things like context managers on top of this API, though perhaps it's smart not to go to far from the wrapped C API, to minimize te maintenance burden as well.

I agree though that this may be a good opportunity to move to a new package. We could start that directly from PROJ 7 and not include the deprecated API. I tried building PROJ 7.0.0 with Yggdrasil, but it needs patches that will be in 7.0.1 that's due next week.

asinghvi17:
Hey, sorry I haven't really looked at this till now.  The code didn't work when I called the low-level transform with vectors, so I made this PR.  With the new API, though, it looks like, e.g., CoordinateTransformations will be a lot easier to use now, so this can probably be closed.

asinghvi17:
Hmm, it looks like the quickstart doesn't work for me.  Seems that the API expects a `Ptr{PJ_AREA}`...

visr:
Oh yeah sorry that was a local patch. You can pass C_NULL there though if you don't use it.

PR #44, opened 2020-06-02
CoordinateTransformations.jl interface through new API

asinghvi17:
This PR implements the CoordinateTransformations interface through a new type `CRS2CRS`, constructed as:
```julia
cs = CRS2CRS("+proj=latlon", "+proj=moll")
# or
source_proj = Projection("+proj=latlon")
dest_proj = Projection("+proj_moll")
cs = CRS2CRS(source_proj, dest_proj)
# Inverse transformations also exist
cs = CRS2CRS("+proj=latlon", "+proj=moll", Proj4.PJ_INV)
# and existing transformations can be trivially inverted by
Base.inv(cs)

# CRS2CRS is callable
using GeoMakie
cs = CRS2CRS("+proj=latlon", "+proj=moll")
lines(cs.(GeoMakie.COASTLINES_LINEVEC))
```
transform_func
There are also some typo fixes, and `proj_coord` has all of its arguments marked as optional.

I'll add some docs once we agree on the API.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -14,7 +14,8 @@ PROJ_jll = "6.2.1, 7"
 julia = "1.3"
 
 [extras]
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Test"]
+test = ["Test", "StaticArrays"]
--- a/Project.toml
+++ b/Project.toml
@@ -9,7 +9,7 @@ PROJ_jll = "58948b4f-47e0-5654-a9ad-f609743f8632"
 
 [compat]
 CEnum = "0.2, 0.3, 0.4"
-CoordinateTransformations = "0.5"
+CoordinateTransformations = "0.5, 0.6"
 PROJ_jll = "6.2.1, 7"
 julia = "1.3"
 
--- a/Project.toml
+++ b/Project.toml
@@ -4,10 +4,12 @@ version = "0.7.2"
 
 [deps]
 CEnum = "fa961155-64e5-5f13-b03f-caf6b980ea82"
+CoordinateTransformations = "150eb455-5306-5404-9cee-2592286d6298"
 PROJ_jll = "58948b4f-47e0-5654-a9ad-f609743f8632"
 
 [compat]
 CEnum = "0.2, 0.3, 0.4"
+CoordinateTransformations = "0.5"
 PROJ_jll = "6.2.1, 7"
 julia = "1.3"
 
--- a/src/coordinate_transformations.jl
+++ b/src/coordinate_transformations.jl
@@ -5,10 +5,6 @@ end
 
 function CRS2CRS(source::String, target::String, direction::PJ_DIRECTION = PJ_FWD; area = C_NULL, normalize = true)
 
-    if direction == PJ_IDENT || source == direction
-        return IdentityTransformation()
-    end
-
     pj_init = proj_create_crs_to_crs(source, target, area)
 
     obj = if normalize
@@ -16,7 +12,7 @@ function CRS2CRS(source::String, target::String, direction::PJ_DIRECTION = PJ_FW
         proj_destroy(pj_init)
         CRS2CRS(pj_normalized, direction)
     else
-        CRS2CRS(pj_normalized, direction)
+        CRS2CRS(pj_init, direction)
     end
 
     finalizer(obj) do obj
@@ -51,7 +47,7 @@ end
 function CoordinateTransformations.transform_deriv(cs::CRS2CRS, x)
     coord = proj_coord(x...)
     factors = proj_factors(cs.rep, coord)
-    return [dx_dlam dx_dphi; dy_dlam dy_dphi]
+    return [factors.dx_dlam factors.dx_dphi; factors.dy_dlam factors.dy_dphi]
 end
 
 function (transformation::CRS2CRS)(x::T) where T
--- a/src/coordinate_transformations.jl
+++ b/src/coordinate_transformations.jl
@@ -53,6 +53,19 @@ end
 
 Base.inv(cs::CRS2CRS) = CRS2CRS(cs.rep, inv(cs.direction))
 
+function Base.show(io::IO, ::MIME"text/plain", cs::CRS2CRS)
+    # info = proj_pj_info(cs.rep)
+    direction = if cs.direction == PJ_FWD
+        "Forward"
+    elseif cs.direction == PJ_INV
+        "Reverse"
+    else
+        "Identity"
+    end
+    println(io, "$(direction)-mode CRS2CRS:")
+    println(io, Proj4.proj_as_proj_string(cs.rep, Proj4.PJ_PROJ_4))
+end
+
 function Base.inv(dir::PJ_DIRECTION)
     return if dir == PJ_IDENT
         PJ_IDENT
--- a/src/coordinate_transformations.jl
+++ b/src/coordinate_transformations.jl
@@ -69,7 +69,7 @@ function CoordinateTransformations.transform_deriv(cs::CRS2CRS, x)
     return [factors.dx_dlam factors.dx_dphi; factors.dy_dlam factors.dy_dphi]
 end
 
-function (transformation::CRS2CRS)(x::T) where T
+function (transformation::CRS2CRS)(x)
     coord = if length(x) == 2
         proj_coord(x[1], x[2],    0,    0)
     elseif length(x) == 3
--- a/src/coordinate_transformations.jl
+++ b/src/coordinate_transformations.jl
@@ -16,6 +16,11 @@ cs = CRS2CRS("+proj=lonlat", "+proj=wintri")
 a = proj_coord(0, 0)
 b = cs(a) # projected coordinate
 ```
+## Extended help
+
+You can get more information on this by calling `Proj4.proj_pj_info(cs.rep)` for a `CRS2CRS` object `cs`.  This will return a `PJ_FACTORS` struct, which you can inspect using `unsafe_string` on its fields.  You can also get a WKT representation
+by calling `Proj4.proj_as_wkt(cs)`
+
 """
 mutable struct CRS2CRS <: CoordinateTransformations.Transformation
     rep::Ptr{Cvoid}
@@ -51,21 +56,6 @@ function CRS2CRS(source::Projection, target::Projection, direction::PJ_DIRECTION
     )
 end
 
-Base.inv(cs::CRS2CRS) = CRS2CRS(cs.rep, inv(cs.direction))
-
-function Base.show(io::IO, ::MIME"text/plain", cs::CRS2CRS)
-    # info = proj_pj_info(cs.rep)
-    direction = if cs.direction == PJ_FWD
-        "Forward"
-    elseif cs.direction == PJ_INV
-        "Reverse"
-    else
-        "Identity"
-    end
-    println(io, "$(direction)-mode CRS2CRS:")
-    println(io, Proj4.proj_as_proj_string(cs.rep, Proj4.PJ_PROJ_4))
-end
-
 function Base.inv(dir::PJ_DIRECTION)
     return if dir == PJ_IDENT
         PJ_IDENT
@@ -75,6 +65,7 @@ function Base.inv(dir::PJ_DIRECTION)
         PJ_FWD
     end
 end
+Base.inv(cs::CRS2CRS) = CRS2CRS(cs.rep, inv(cs.direction))
 
 function CoordinateTransformations.transform_deriv(cs::CRS2CRS, x)
     coord = proj_coord(x...)
@@ -83,15 +74,10 @@ function CoordinateTransformations.transform_deriv(cs::CRS2CRS, x)
 end
 
 function (transformation::CRS2CRS)(x)
-    coord = if length(x) == 2
-        PJ_COORD(x[1], x[2],    0,    0)
-    elseif length(x) == 3
-        PJ_COORD(x[1], x[2], x[3],    0)
-    elseif length(x) == 4
-        PJ_COORD(x[1], x[2], x[3], x[4])
-    else
-        error("Input must have length 2, 3 or 4! Found $(length(x)).")
-    end
+    # if `x` is too long, this will throw
+    # a methoderror (there are 0, 1, 2, 3, and 4-arg)
+    # versions of the PJ_COORD constructor.
+    coord = PJ_COORD(x...)
 
     xyzt = (proj_trans(transformation.rep, transformation.direction, coord).xyzt)::PJ_XYZT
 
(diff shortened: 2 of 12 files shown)


asinghvi17:
A benchmark on a length-5000 array shows:
```julia
BenchmarkTools.Trial: 
  memory estimate:  1.00 MiB
  allocs estimate:  42100
  --------------
  minimum time:     2.783 ms (0.00% GC)
  median time:      3.014 ms (0.00% GC)
  mean time:        3.234 ms (5.29% GC)
  maximum time:     14.702 ms (77.94% GC)
  --------------
  samples:          1545
  evals/sample:     1
```
so I don't know whether it's worth it to add custom broadcasting support for this (though it could certainly be done).

src/coordinate_transformations.jl
```diff
@@ -0,0 +1,91 @@
...
+        proj_destroy(obj.rep)
+    end
+
+    return obj
+end
+
+function CRS2CRS(source::Projection, target::Projection, direction::PJ_DIRECTION = PJ_FWD; area = C_NULL, normalize = true)
+    return CRS2CRS(
+        sprint(print, source),
+        sprint(print, target),
```
c42f, line 32: Rather than using `sprint` here, I see that there's `proj_create_crs_to_crs_from_pj` which might be useful?

Project.toml
```diff
@@ -4,10 +4,12 @@ version = "0.7.2"
 
 [deps]
 CEnum = "fa961155-64e5-5f13-b03f-caf6b980ea82"
+CoordinateTransformations = "150eb455-5306-5404-9cee-2592286d6298"
 PROJ_jll = "58948b4f-47e0-5654-a9ad-f609743f8632"
 
 [compat]
 CEnum = "0.2, 0.3, 0.4"
+CoordinateTransformations = "0.5"
```
c42f, line 12: Please add 0.6 as well which is out recently.

src/coordinate_transformations.jl
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    pj_init = proj_create_crs_to_crs(source, target, area)
+
+    obj = if normalize
+        pj_normalized = proj_normalize_for_visualization(pj_init)
+        proj_destroy(pj_init)
+        CRS2CRS(pj_normalized, direction)
+    else
+        CRS2CRS(pj_normalized, direction)
```
c42f, line 19: pj_normalize doesn't exist here?

src/coordinate_transformations.jl
```diff
@@ -0,0 +1,91 @@
+mutable struct CRS2CRS <: CoordinateTransformations.Transformation
+    rep::Ptr{Cvoid}
+    direction::PJ_DIRECTION
+end
+
+function CRS2CRS(source::String, target::String, direction::PJ_DIRECTION = PJ_FWD; area = C_NULL, normalize = true)
+
+    if direction == PJ_IDENT || source == direction
+        return IdentityTransformation()
```
c42f, line 9: Just a note that this makes `CRS2CRS` a little type unstable. I'm not sure whether that's worth it or not?

src/coordinate_transformations.jl
```diff
@@ -0,0 +1,91 @@
...
+        PJ_INV
+    else
+        PJ_FWD
+    end
+end
+
+function CoordinateTransformations.transform_deriv(cs::CRS2CRS, x)
+    coord = proj_coord(x...)
+    factors = proj_factors(cs.rep, coord)
+    return [dx_dlam dx_dphi; dy_dlam dy_dphi]
```
c42f, line 54: dx_dlam etc doesn't exist?

src/coordinate_transformations.jl
```diff
@@ -0,0 +1,91 @@
...
+end
+
+function (transformation::CRS2CRS)(x::T) where T
```
c42f, line 57: Is this `where T` needed?  Seems it could be removed.

src/coordinate_transformations.jl
```diff
@@ -0,0 +1,91 @@
+mutable struct CRS2CRS <: CoordinateTransformations.Transformation
+    rep::Ptr{Cvoid}
+    direction::PJ_DIRECTION
+end
+
+function CRS2CRS(source::String, target::String, direction::PJ_DIRECTION = PJ_FWD; area = C_NULL, normalize = true)
```
c42f, line 6: I don't think `normalize=true` should be the default if it results in units which are not standard.

src/coordinate_transformations.jl
asinghvi17: I tried that, but got pretty consistent segfaults.  Not sure why...

src/coordinate_transformations.jl
asinghvi17: Yeah, I wasn't sure about this myself.  We could definitely get rid of it.

src/coordinate_transformations.jl
asinghvi17: AFAIK, the only thing normalize does is change the axis order, but I might have missed something in the docs.  I got that info [here](https://proj.org/development/reference/functions.html?highlight=normalize_for_visualization#c.proj_normalize_for_visualization).

src/coordinate_transformations.jl
asinghvi17: Would the method still specialize on types with size information (e.g. static arrays) then?

asinghvi17:
To address @visr's comment on Slack,

The reasons I chose to adopt CoordinateTransformations are:
-  it seems to be a pretty standard interface which works generically
- we're thinking of adding some more "first-class" support for it in Makie, in the future
- there are clear ways to get derivatives and inverses, which are also available through the Proj interface
- even without the CoordinateTransformations interface, the interface would have looked like this anyway, since that seems to be what Proj is aiming towards.

c42f:
I'd point out that way back when @andyferris and I were talking about the initial design of CoordinateTransformations we were very much thinking about geospatial, and that as a consequence Geodesy.jl also uses the CoordinateTransformations types (albeit not well maintained right now, I hope Geodesy.jl will eventually get a new caring maintainer :-) )

The great benefit of having the transformation behave like a normal function is that it naturally works in any place that a function is expected.

I believe Proj also provides a way to compose transformations into a pipeline? In that case it may also be worth implementing `CoordinateTransformations.compose(::CRS2CRS, ::CRS2CRS)` on top of that? In principle Proj may be able to optimize the pipeline to be more accurate or faster.

src/coordinate_transformations.jl
c42f: Ugh, [that trick](https://docs.julialang.org/en/v1/manual/performance-tips/index.html#Be-aware-of-when-Julia-avoids-specializing-1) again! According to the docs it's only important for inputs of type `Function`, `Type` and `Vararg`. So I guess not needed here; I'd delete it, unless you can measure a performance difference.

src/overloads.jl
```diff
@@ -0,0 +1,29 @@
+# Convenience functions for PJ_COORD
+
+Base.length(::PJ_COORD) = 4
+
+Base.@propagate_inbounds function Base.getindex(coord::PJ_COORD, i::Int)
```
c42f, line 5: Rather than doing this, you could perhaps consider making `PJ_COORD` into a `FieldVector` subtype. Then it becomes a proper `StaticArray`. Would require tweaking the clang-based code generator.

c42f: Though having said that, I'm not 100% sure that it's desired to have all array operations defined on this. PJ_COORD isn't really a mathematical vector at all.

src/proj_c.jl
```diff
@@ -104,7 +104,7 @@ function proj_trans_generic(P, direction, x, sx, nx, y, sy, ny, z, sz, nz, t, st
     ccall((:proj_trans_generic, libproj), Csize_t, (Ptr{PJ}, PJ_DIRECTION, Ptr{Cdouble}, Csize_t, Csize_t, Ptr{Cdouble}, Csize_t, Csize_t, Ptr{Cdouble}, Csize_t, Csize_t, Ptr{Cdouble}, Csize_t, Csize_t), P, direction, x, sx, nx, y, sy, ny, z, sz, nz, t, st, nt)
 end
 
-function proj_coord(x, y, z, t)
+function proj_coord(x = 0, y = 0, z = 0, t = 0)
     ccall((:proj_coord, libproj), PJ_COORD, (Cdouble, Cdouble, Cdouble, Cdouble), x, y, z, t)
```
c42f, line 108: Should we even be calling `proj_coord`? Seems kinda wasteful use of a `ccall` if all it does is fill in the `PJ_COORD` element of a union when we could call the PJ_COORD constructor directly at the Julia level.

src/overloads.jl
asinghvi17: Can't it be thought of as a 4-vector (or, at least, a 4-point)?

c42f: It's a space time 4-point, yes, but generally has no vector properties or even affine space properties (so we're back to the usual LinearAlgebra-object vs array abstraction problem). It's certainly an array of components though so making it an `AbstractArray` would be no more terrible than other non-vector things which get stored in `Vector` :-)

visr:
I updated the `gen/wrap_proj.jl` to automatically make the changes you made to the generated files. Shall I push that onto this branch? It will also directly bring in the new functions defined in PROJ 7, such as `proj_create_crs_to_crs_from_pj`.

asinghvi17:
Thanks, @visr!  Unfortunately, my laptop just broke down, so I'm migrating to an old one we had in the house.  It's going to take a while before I can make any major changes to this :D - but IMO, after some cleanup, this should be good to go.

asinghvi17:
With this latest commit, the Base.show method has been overridden to :
```julia
julia> CRS2CRS("+proj=lonlat", "+proj=wintri")
Forward-mode CRS2CRS:
+proj=pipeline +step +proj=unitconvert +xy_in=deg +xy_out=rad +step +proj=wintri +lon_0=0 +lat_1=0 +x_0=0 +y_0=0 +ellps=WGS84
```

visr:
Nice! However I don't think we should be promoting proj strings that much, given that proj docs themselves state that generally WKT or SRID should be used, so that is probably a better candidate for show: https://proj.org/faq.html#what-is-the-best-format-for-describing-coordinate-reference-systems

asinghvi17:
Hmm....I wonder if I should drop the custom show, then - WKT would be pretty long if I remember correctly.

src/coordinate_transformations.jl
c42f: Oh I see. Well I still think that probably shouldn't be the default. The CRS contains metadata specifying the ordering of components (lat,lon) / (lon,lat) / (northing,easting) / (easting,northing), so asking for a specific CRS is asking for that ordering and it should probably be respected.

Granted there's a tension here between analytical work where normalize=false should be preferred, vs plotting where convenience is arguably more important. But I think as a Proj API it should be correct rather than convenient by default :-)

c42f:
WKT is definitely preferable for serialization. It would be interesting to compare and see whether it's usable with `show` or is just too verbose for interactive use.

src/overloads.jl
```diff
@@ -0,0 +1,33 @@
...
+
+    return if i == 1
+        coord.xyzt.x
+    elseif i == 2
+        coord.xyzt.y
+    elseif i == 3
+        coord.xyzt.x
+    elseif i == 4
+        coord.xyzt.t
+    end
```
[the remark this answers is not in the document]
asinghvi17: Cool, didn't know about that!  Will make the change.

src/coordinate_transformations.jl
```diff
@@ -0,0 +1,98 @@
...
+function (transformation::CRS2CRS)(x)
+    coord = if length(x) == 2
+        PJ_COORD(x[1], x[2],    0,    0)
+    elseif length(x) == 3
+        PJ_COORD(x[1], x[2], x[3],    0)
+    elseif length(x) == 4
+        PJ_COORD(x[1], x[2], x[3], x[4])
+    else
+        error("Input must have length 2, 3 or 4! Found $(length(x)).")
+    end
```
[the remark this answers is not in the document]
asinghvi17: It will throw a MethodError now, but does require the coordinate type to be iterable.  I'm not sure whether that's a problem, though.

src/coordinate_transformations.jl
```diff
@@ -0,0 +1,98 @@
...
+
+    xyzt = proj_trans(transformation.rep, transformation.direction, coord).xyzt
+
+    return if length(x) == 2
+        T(xyzt.x, xyzt.y)
+    elseif length(x) == 3
+        T(xyzt.x, xyzt.y, xyzt.z)
+    elseif length(x) == 4
+        T(xyzt.x, xyzt.y, xyzt.z, xyzt.t)
```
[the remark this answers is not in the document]
asinghvi17: This already works with GeometryBasics, but I will test with StaticArrays directly.

src/coordinate_transformations.jl
```diff
@@ -0,0 +1,98 @@
...
+        PJ_INV
+    else
+        PJ_FWD
+    end
+end
+
+function CoordinateTransformations.transform_deriv(cs::CRS2CRS, x)
+    coord = proj_coord(x...)
+    factors = proj_factors(cs.rep, coord)
+    return [factors.dx_dlam factors.dx_dphi; factors.dy_dlam factors.dy_dphi]
```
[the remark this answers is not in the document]
asinghvi17: Makes sense to get rid of it, then (especially since there seems to be no z-transform).

src/coordinate_transformations.jl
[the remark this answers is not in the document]
asinghvi17: Ah, I accidentally removed the `x::T where T` earlier, have added that back in.  It does work with StaticArrays, btw - the idea was that it can work for anything which has a constructor of this form, which is most coordinate types.

src/proj_c.jl
```diff
@@ -26,11 +50,199 @@ function proj_context_get_use_proj4_init_rules(from_legacy_code_path, ctx = C_NU
     ccall((:proj_context_get_use_proj4_init_rules, libproj), Cint, (Ptr{PJ_CONTEXT}, Cint), ctx, from_legacy_code_path)
 end
 
+"""
+    proj_context_set_fileapi(const PROJ_FILE_API * fileapi,
```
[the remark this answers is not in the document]
visr: This is basically just rerunning the existing wrapping scripts. Would be good to have a look. But I realize it blows up the diff, so I can also take it out if you prefer and make a separate PR.

visr:
Thanks for the reviews, Chris.

> I'm still skeptical about normalize=true. I'm not sure the library should make this choice for users. I'm not entirely certain though and would like to hear other opinions about this. What do you think @visr?

Indeed difficult. Given this is a PROJ wrapper library we should probably have strong reasons to deviate from default PROJ behavior, so my inclination is to `normalize=false`. It's probably bound to create confusion either way. Most point types in use are just two ordered coordinates, with no way to specify what is what. In ArchGDAL we ended up going the same route, see https://github.com/yeesian/ArchGDAL.jl/issues/119.

> If you really must convert to string to represent CRS, WKT should be preferred
I think making PJ_COORD into an AbstractVector would help things. If @visr is willing to take an explicit StaticArrays dependency, the easiest way forward here may simply be to use PJ_COORD <: FieldVector{4,Float64}

This is fine by me.

I should check, but perhaps for printing we can first check if it is a SRID (like EPSG), and if so, print that, otherwise WKT.

c42f:
> I should check, but perhaps for printing we can first check if it is a SRID (like EPSG), and if so, print that, otherwise WKT.

That sounds like a good idea for pretty printing.

What I'm more concerned about is the use of `string(::Projection)` in the `CRS2CRS(::Projection, ::Projection, ...)` constructor. (This should really be avoided, but if absolutely necessary we should go via WKT.)

src/proj_c.jl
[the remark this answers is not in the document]
c42f: Oh I see. Might be nice to merge it separately if that's easy.

visr:
> What I'm more concerned about is the use of string(::Projection) in the CRS2CRS(::Projection, ::Projection, ...) constructor.

I agree. I got it to work with `proj_create_crs_to_crs_from_pj` you suggested [above](https://github.com/JuliaGeo/Proj4.jl/pull/44#discussion_r434399530) in c1d9e634af5b51fdf55ff380681ba1b7267d3013, which I committed in a separate branch. But going down that path actually led me to believe we should probably just remove the `CRS2CRS(::Projection, ::Projection, ...)` constructor.

CRS2CRS is thankfully built on the new API. But this constructor ties it to the old API, since `Projection` is a higher level construct based on the old API. In my commit I tried to move `Projection` over to `proj_create`, but this led to so much breakage that I stopped.

c42f:
> But going down that path actually led me to believe we should probably just remove the CRS2CRS(::Projection, ::Projection, ...) constructor.

That's probably a better option. `Projection` was always a broken abstraction given that it's kind of a CRS but kind of a transformation.

visr:
Perhaps the best way forward here is to start taking bits of this, and get them merged in parts? The PROJ 7 wrapping was already on master, I forgot I had already done that here, and I just submitted another bitesize PR from this in #49.

c42f:
Sounds great.

visr:
Closed by #51. Thanks a lot for starting the effort with this PR!

PR #51, opened 2021-01-27
CoordinateTransformations.jl API

visr:
~~Not yet finished, but parking the work in progress PR here until I have time to finish it.~~
Part of #44, see https://github.com/JuliaGeo/Proj4.jl/pull/44#issuecomment-759693533

### TODO
- [x] documentation (input can be any indexable length 2-4, output is same length, but SVector{Float64}), network demo (quickstart in README done)
- [x] Base.show(::Transformation)
- [x] Add methods for vectors of points, use proj_trans_array / proj_trans_generic? (better in a separate PR, if at all)
- [x] Implement `CoordinateTransformations.compose`

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,8 @@ trans([55, 12])
 
 # Note that above the latitude is passed first, because that is the axis order that the
 # EPSG mandates. If you want to pass in (longitude, latitude) / (x, y), you can set the
-# `normalize` keyword to true. For more info see https://proj.org/faq.html#why-is-the-axis-ordering-in-proj-not-consistent
-trans = Proj4.Transformation("EPSG:4326", "+proj=utm +zone=32 +datum=WGS84", normalize=true)
+# `always_xy` keyword to true. For more info see https://proj.org/faq.html#why-is-the-axis-ordering-in-proj-not-consistent
+trans = Proj4.Transformation("EPSG:4326", "+proj=utm +zone=32 +datum=WGS84", always_xy=true)
 
 # now we input (longitude, latitude), and get the same result as above
 trans([12, 55])
--- a/README.md
+++ b/README.md
@@ -30,9 +30,9 @@ trans([12, 55])
 
 # using `inv` we can reverse the direction, `compose` can combine two transformations in one
 inv(trans)([691875.632, 6098907.825]) ≈ [12, 55]
+```
 
-
-
+```julia
 # This is the old API of this package, which will be removed soon
 wgs84 = Projection("+proj=longlat +datum=WGS84 +no_defs")
 utm56 = Projection("+proj=utm +zone=56 +south +datum=WGS84 +units=m +no_defs")
--- a/README.md
+++ b/README.md
@@ -4,11 +4,36 @@
 
 A simple Julia wrapper around the [PROJ](https://proj.org/) cartographic projections library.
 
-Basic example:
+Quickstart, based on the [PROJ docs](https://proj.org/development/quickstart.html):
 
 ```julia
 using Proj4
 
+# Proj4.jl implements the CoordinateTransformations.jl API
+# A Proj4.Transformation needs the source and target coordinate reference systems
+trans = Proj4.Transformation("EPSG:4326", "+proj=utm +zone=32 +datum=WGS84")
+
+# Once created, you can call this object to transform points
+# the result will be a SVector from StaticArrays.jl
+# Here the (latitude, longitude) of Copenhagen is entered
+trans([55, 12])
+# -> SVector{2, Float64}(691875.632, 6098907.825)
+
+# Note that above the latitude is passed first, because that is the axis order that the
+# EPSG mandates. If you want to pass in (longitude, latitude) / (x, y), you can set the
+# `normalize` keyword to true. For more info see https://proj.org/faq.html#why-is-the-axis-ordering-in-proj-not-consistent
+trans = Proj4.Transformation("EPSG:4326", "+proj=utm +zone=32 +datum=WGS84", normalize=true)
+
+# now we input (longitude, latitude), and get the same result as above
+trans([12, 55])
+# -> SVector{2, Float64}(691875.632, 6098907.825)
+
+# using `inv` we can reverse the direction, `compose` can combine two transformations in one
+inv(trans)([691875.632, 6098907.825]) ≈ [12, 55]
+
+
+
+# This is the old API of this package, which will be removed soon
 wgs84 = Projection("+proj=longlat +datum=WGS84 +no_defs")
 utm56 = Projection("+proj=utm +zone=56 +south +datum=WGS84 +units=m +no_defs")
 
--- a/README.md
+++ b/README.md
@@ -41,5 +41,12 @@ transform(wgs84, utm56, [150 -27 0])
 # Should result in [202273.913 7010024.033 0.0]
 ```
 
+Note that, as described in https://proj.org/resource_files.html, PROJ has the capability
+to use remote grids for transformations that need them. For users that have not set
+the environment variable `PROJNETWORK=ON`, or changed `proj.ini`, the networking is
+disabled by default. To enable from Julia, run `Proj4.enable_network()`.
+`Proj4.network_enabled()` can be used to check the setting. Note that it needs to be set
+before creating a transformation, otherwise it will have no effect.
+
 API documentation for the underlying C API may be found here:
 https://proj.org/development/reference/index.html
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ transform(wgs84, utm56, [150 -27 0])
 Note that, as described in https://proj.org/resource_files.html, PROJ has the capability
 to use remote grids for transformations that need them. For users that have not set
 the environment variable `PROJNETWORK=ON`, or changed `proj.ini`, the networking is
-disabled by default. To enable from Julia, run `Proj4.enable_network()`.
+disabled by default. To enable from Julia, run `Proj4.enable_network!()`.
 `Proj4.network_enabled()` can be used to check the setting. Note that it needs to be set
 before creating a transformation, otherwise it will have no effect.
 
--- a/src/coord.jl
+++ b/src/coord.jl
@@ -1,1 +1,53 @@
 const Coord = SVector{4, Float64}
+
+mutable struct Transformation <: CoordinateTransformations.Transformation
+    pj::Ptr{PJ}
+    function Transformation(pj::Ptr{PJ})
+        tr = new(pj)
+        finalizer(tr) do tr
+            tr.pj = proj_destroy(tr.pj)
+        end
+        return tr
+    end
+end
+
+function Transformation(
+    source_crs::AbstractString,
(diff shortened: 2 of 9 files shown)


src/coord.jl
```diff
@@ -1,17 +1,81 @@
-"""
-    Coord <: FieldVector{4, Float64}
+const Coord = SVector{4, Float64}
```
c42f, line 1: Yeah this probably makes sense. What drew you to this conclusion?

visr: Mostly it felt wrong that the field names were not necessarily correct, for instance x and y could be reversed. I wrote a bit more detail in the commit message: https://github.com/JuliaGeo/Proj4.jl/pull/51/commits/1d0c03ef93701a2f4a2d36d8ea3eea0238e152d1

c42f: Ah yes I forgot about CRS-dependent axis order. That's a very good reason to go with indices only.

visr:
I'm looking for some input on input, related to this bit of code:
https://github.com/JuliaGeo/Proj4.jl/blob/366a5ff4046bcf73564ae51815d51160d868bd30/src/coord.jl#L74-L112

The PROJ C API always works with length 4 arrays, where the first two positions are x or y, then height and time. For when you only care about the x and y, we default to a height of 0 and infinite time, just like the `cs2cs` application. In `cs2cs`, if your input is size 4, you get size 4 back. If the input is size 2 or 3, you get size 3 back, always including height. My idea was to instead always pass back the same size as your input. If you care about height, you will only get it back if you pass it in as well. Does that sound reasonable? Or is it better to follow `cs2cs` here, and always return height, such that users are more aware if the operation affects the height as well.


Besides the length, there is also the matter what types to accept and return. Right now, I always return `SVector{Float64}`, with length as above. If the input is an `SVector` as well, I can specialize on the length, otherwise, for e.g. `AbstractVector`/`Tuple` input I check the length and adjust the output length accordingly. Though I guess that would mean it is type unstable. Is this the right approach? Any other ideas?

Alternatively I could try to give back the same type that comes in, but with eltype `Float64`. E.g. people passing tuples might prefer a tuple back, and people passing a `GeometryBasics.Point` might prefer that back, using a method like `f(x) = typeof(x)(SA_F64[x...])` which can return that. Or would this make things too complicated?

Would be happy to get ideas from anyone, tagging @c42f since he helped out before.

c42f:
> My idea was to instead always pass back the same size as your input. If you care about height, you will only get it back if you pass it in as well. Does that sound reasonable?

Sounds reasonable to me. As an extension of this, you could check that any dummy time and height values passed are returned back unchanged. If not, throw an error. (Optionally have a flag in the transformation to turn off such checking if performance concerns warrant it?)

> If the input is an SVector as well, I can specialize on the length, otherwise, for e.g. AbstractVector/Tuple input I check the length and adjust the output length accordingly. Though I guess that would mean it is type unstable. Is this the right approach? Any other ideas?

I think the right approach here is to use `StaticArrays.similar_type` for any input which is a subtype of `StaticVector` (ie, generalize your functions which currently take only `SVector`)

For other `AbstractVector`, either:

* Type instability may not be so bad if the compiler can figure out that the return type is `Union{SVector{2.Float64},SVector{3.Float64},SVector{4.Float64}}`? That way the small union optimization may kick in. The compiler can probably figure this out but you could add a type assertion if necessary.  Do a performance test?
* Failing that, you should probably just use `Base.similar()` to construct an appropriate array which can be filled.

visr:
Thanks for the helpful tips! I tried to implement the changes in d60567c17d722b4fe8fcbd504286c9deaac7604a.

> As an extension of this, you could check that any dummy time and height values passed are returned back unchanged. If not, throw an error.

I'm afraid this would get in the way for 2D applications that don't mind also if the height is changed. I believe the time always comes back unchanged.

> Type instability may not be so bad if the compiler can figure out that the return type is Union{SVector{2,Float64},SVector{3,Float64},SVector{4,Float64}}?

For a size 2 `Vector{Float64}` input, running with `Test.@inferred` I get

`SVector{2, Float64} does not match inferred return type SArray{S, Float64, 1, L} where {S<:Tuple, L}`.
Since `SVector{2, Float64}` is an alias for `SArray{Tuple{2}, Float64, 1, 2})`, I guess this means the compiler did not figure out that it is 2, 3, or 4, but any `n`? Not sure if that is worse than a `Union` of 3 types.

BenchmarkTools shows it is almost equally fast (1.3 μs) for SVector, Vector and Tuple input. Only for Vector it gives a single allocation. So I guess that is good enough.

README.md
```diff
@@ -4,11 +4,36 @@
...
+# Once created, you can call this object to transform points
+# the result will be a SVector from StaticArrays.jl
+# Here the (latitude, longitude) of Copenhagen is entered
+trans([55, 12])
+# -> SVector{2, Float64}(691875.632, 6098907.825)
+
+# Note that above the latitude is passed first, because that is the axis order that the
+# EPSG mandates. If you want to pass in (longitude, latitude) / (x, y), you can set the
+# `normalize` keyword to true. For more info see https://proj.org/faq.html#why-is-the-axis-ordering-in-proj-not-consistent
+trans = Proj4.Transformation("EPSG:4326", "+proj=utm +zone=32 +datum=WGS84", normalize=true)
```
c42f, line 25: I think `normalize` could be a confusing name here, as there's already `LinearAlgebra.normalize()` which means something very different.

Possible alternative might be `ordering=:latlong` vs `ordering=:longlat`?  `ordering` could be `axis_order`? or if you want the bool, `reorder_axes=true` perhaps?

c42f:
> I'm afraid this would get in the way for 2D applications that don't mind also if the height is changed.

True. In that case I guess what you've got is fine.

> I believe the time always comes back unchanged.

Ah of course... it would be surprising if proj supported nontrivial temporal reference frames / datum shifts :laughing: 

> SVector{2, Float64} does not match inferred return type SArray{S, Float64, 1, L} where {S<:Tuple, L}.
Since SVector{2, Float64} is an alias for SArray{Tuple{2}, Float64, 1, 2}), I guess this means the compiler did not figure out that it is 2, 3, or 4, but any n? Not sure if that is worse than a Union of 3 types.

Ok, good to know. In that case, I'd suggest adding a type assertion with the Union of three types and checking whether that helps. You'll also want to benchmark actually doing something with the returned value (eg, something simple such as accumulating it with `+`), because that's where the performance hit will be. I think a union of three types should be better than the `UnionAll` which the compiler is inferring.

README.md
visr: Yeah good point. I would prefer a boolean option. Actually I checked what pyproj uses, and I quite like their `always_xy`, 1b41e863186d4277cfe172343926f6e77c29c6e9. How's that?

visr:
> I'd suggest adding a type assertion with the Union of three types and checking whether that helps

It doesn't seem to help. I updated the benchmark, and the performance of vectors is really quite close, so perhaps it's fine to leave it. This is the benchmark I used:

```julia
using StaticArrays
using BenchmarkTools
using Proj4
using Test

const trans = Proj4.Transformation("EPSG:4326", "EPSG:28992", always_xy = true)

# test different input types
ps = SA[5.39, 52.16]
pv = [5.39, 52.16]
pt = (5.39, 52.16)

function timer(p)
    tot = 0.0
    for _ in 1:10_000
        tot += sum(trans(p))
    end
    tot
end

@btime timer($ps) # 13.512 ms (0 allocations: 0 bytes)
@btime timer($pv) # 16.543 ms (30000 allocations: 625.00 KiB)
@btime timer($pt) # 13.179 ms (0 allocations: 0 bytes)

Test.@inferred trans(ps)  # pass
Test.@inferred trans(pv)  # return type SVector{2, Float64} does not match inferred return type SArray{S, Float64, 1, L} where {S<:Tuple, L}
Test.@inferred trans(pt)  # pass
```

Overall I think this PR is ready. Does anyone have other feedback or is this good to merge? Will merge and tag in a few days without comments.

src/coord.jl
```diff
@@ -1 +1,157 @@
...
+
+"""
+    enable_network(active::Bool = true, ctx::Ptr{PJ_CONTEXT} = C_NULL)::Bool
+
+Enable PROJ network access, if `active` is true, disable it if it is false. Optionally pass
+a context to set it for that context, instead of the global one.
+
+Returns true if network access is possible.
+"""
+function enable_network(active::Bool = true, ctx::Ptr{PJ_CONTEXT} = C_NULL)
```
yeesian, line 143: (optional) consider breaking this out into two functions to prevent any risk of confusion: `enable_network!(...)` and `disable_network!(...)`.

src/coord.jl
```diff
@@ -1 +1,157 @@
...
+    function Transformation(pj::Ptr{PJ})
+        trans = new(pj)
+        finalizer(trans) do trans
+            trans.pj = proj_destroy(trans.pj)
+        end
+        return trans
+    end
+end
+
+function Transformation(
```
yeesian, line 15: it'll be nice if this function has a docstring, maybe describing (or giving examples of) the kinds of strings that are valid for `source_crs` and `target_crs`.

README.md
```diff
@@ -4,17 +4,49 @@
...
+trans = Proj4.Transformation("EPSG:4326", "+proj=utm +zone=32 +datum=WGS84", always_xy=true)
+
+# now we input (longitude, latitude), and get the same result as above
+trans([12, 55])
+# -> SVector{2, Float64}(691875.632, 6098907.825)
+
+# using `inv` we can reverse the direction, `compose` can combine two transformations in one
+inv(trans)([691875.632, 6098907.825]) ≈ [12, 55]
+
+
```
yeesian, line 34: redundant whitespace?

yeesian (approved):
Drive-by review: looks great!

README.md
c42f: Sure, seems ok to me.

c42f:
> It doesn't seem to help. I updated the benchmark, and the performance of vectors is really quite close, so perhaps it's fine to leave it. This is the benchmark I used:

In that case looks good to me. It seems that the cost of the transformation is just relatively large compared to other costs in this code.

I also tried out `Geodesy.UTMfromLLA` and it was slightly slower than the comparable transformation from Proj so it seems that the cost of interop with the proj C library is well under control here.  (Albeit being slightly disappointing in other aspects :sweat_smile: )

c42f (approved):

src/coord.jl
visr: Yeah good point, see 5212ae01c67bb38e757d69bbe196d4b398f79b5a

src/coord.jl
visr: I think I'd prefer keeping it together in a single function, `enable_network!(false)` meaning `disable_network!()` doesn't seem too strange to me.

But you're right that this needs a !: 625db851d3c20468f3dfa07c3790ad306df14df0

README.md
visr: Yeah wanted to keep them apart, separate code blocks are better: c7b0c522bee9dd46abbb8d567687270d7c3e1e52

c42f:
Nice!

PR #54, opened 2021-08-18
Fix overwriting 4th coordinate in coord.jl

disberd:
I am studying a bit the internal of Proj4.jl to undertand how to exploit it better, but it seem that the current implementation on the new API wrongly overwrites the 4th coordinate if this is provided with a 4 element SVector as input.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Proj4"
 uuid = "9a7e659c-8ee8-5706-894e-f68f43bc57ea"
-version = "0.7.5"
+version = "0.7.6"
 
 [deps]
 CEnum = "fa961155-64e5-5f13-b03f-caf6b980ea82"
--- a/src/coord.jl
+++ b/src/coord.jl
@@ -141,7 +141,7 @@ end
 
 function (trans::Transformation)(coord::StaticVector{4,<:AbstractFloat})
     T = similar_type(coord)
-    coord = SVector{4, Float64}(coord[1], coord[2], coord[3], Inf)
+    coord = SVector{4, Float64}(coord[1], coord[2], coord[3], coord[4])
     p = proj_trans(trans.pj, PJ_FWD, coord)
     return T(p)
 end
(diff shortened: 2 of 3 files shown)


visr:
Thanks! Indeed this looks like an unfortunate copy paste error.

Could you perhaps add a test for this? The last one from this block shows the effect of passing a non-Inf time coordinate:
https://github.com/JuliaGeo/Proj4.jl/blob/b00cad114be5c90a91072eb1a0d96f5b4717d4d0/test/applications.jl#L79-L88

Since you need network for that test, perhaps it is best to add the test here:
https://github.com/JuliaGeo/Proj4.jl/blob/b00cad114be5c90a91072eb1a0d96f5b4717d4d0/test/proj6api.jl#L281

And feel free to already bump the patch version, such that we can tag a release right after.

disberd:
> Thanks! Indeed this looks like an unfortunate copy paste error.
> 
> Could you perhaps add a test for this? The last one from this block shows the effect of passing a non-Inf time coordinate:
> https://github.com/JuliaGeo/Proj4.jl/blob/b00cad114be5c90a91072eb1a0d96f5b4717d4d0/test/applications.jl#L79-L88
> 
> Since you need network for that test, perhaps it is best to add the test here:
> https://github.com/JuliaGeo/Proj4.jl/blob/b00cad114be5c90a91072eb1a0d96f5b4717d4d0/test/proj6api.jl#L281
> 
> And feel free to already bump the patch version, such that we can tag a release right after.

Hi, Thanks for the fast reply.

I tried looking at the test you suggested but I do not understand why you say that one would need the network to perform this kind of test.
Looking at the file, I decided to put this test in the `single transformation` test-set rather than the network and it seems that the tests correctly fail in master but passes in this patched version.
I never did a test before so sorry if I wrongly assumed I did not need network and put the test in the wrong place.

visr:
> I tried looking at the test you suggested but I do not understand why you say that one would need the network to perform this kind of test.

You're right, the test you added also properly tests if the time coordinate is not overwritten. I was thinking about the other test with networking because that is an example of a transformation where the time actually affects the x and y coordinate transformations, to show that it is not only passed along, but also does something. But that isn't really needed to test this fix, your solution is simpler.

PR #63, opened 2022-04-27
Add a direction field to Transformation

asinghvi17:
Pursuant to discussion in #57, this adds a `direction::PJ_DIRECTION` field to the Transformation struct.  It also implements this as a keyword argument in constructors, and implements it in the transformation functions.

### DIFF
--- a/src/Proj.jl
+++ b/src/Proj.jl
@@ -6,6 +6,7 @@ using StaticArrays
 using CoordinateTransformations
 
 export PROJ_jll
+export PJ_DIRECTION, PJ_FWD, PJ_IDENT, PJ_INV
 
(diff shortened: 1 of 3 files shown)


visr (approved):
Thanks! I pushed a commit, 77bd403813154fcb960f30bc818ed38568667cfe. Could you have a look to see if you agree with these, mostly minor, changes?

For usability, I thought to export `PJ_FWD` etc. Also added tests, ran the formatter, restored the `f(; ctx)` keyword syntax, and added support for identity transforms.

asinghvi17:
Thanks!  Makes sense to me, and the code is definitely cleaner.

PR #64, opened 2022-05-03
Remove StaticArrays dependency

visr:
Change `Coord` from being a `SVector{4, Float64}` alias to a struct implemented here. This struct is now mostly internal, since tuples are always returned from transformations. Length of the output tuple still depends on the input length. SVectors still work as input coordinates, just like tuples, vectors, and multiple arguments `f(x, y)`.

I included a benchmark here that shows that performance is good:
https://github.com/JuliaGeo/Proj4.jl/blob/a2e71c159a5892b0986e1b933fe0ff405ce940a2/test/benchmark.jl

This is the same benchmark as used here: https://github.com/JuliaGeo/Proj4.jl/pull/51#issuecomment-864467491. For types like SVector or tuples the return type (tuple length) can be inferred precisely, for Vector it infers `Union{NTuple{2,Float64},NTuple{3,Float64},NTuple{4,Float64}}` which is as good as we can expect.

This is on top of #57, but since it changes the return type of transformations I wanted to put it up separately for review. I think after this #57 would be ready for release, I was thinking to make it Proj.jl v1, such that we can actually start using the minor version number for features.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,7 @@ version = "1.0.0"
 [deps]
 CEnum = "fa961155-64e5-5f13-b03f-caf6b980ea82"
 CoordinateTransformations = "150eb455-5306-5404-9cee-2592286d6298"
+NetworkOptions = "ca575930-c2e3-43a9-ace4-1e988b2c1908"
 PROJ_jll = "58948b4f-47e0-5654-a9ad-f609743f8632"
 
 [compat]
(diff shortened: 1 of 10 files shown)


asinghvi17 (commented):
I think that the change makes sense, and I don't use the Coord interface at all on the GeoMakie end.  If performance is the same then I don't see any reason not to merge, especially if it simplifies the API (and makes it more intuitive for other users).

There seems to be an odd type issue with the Makie internals and GeoMakie going on right now, but aside from that I think this should be good to merge.

visr:
Thanks for having a look, and good point about the return type. Returning the same type was originally a design goal, see also old discussion [here](https://github.com/JuliaGeo/Proj4.jl/pull/51#issuecomment-851748327). It looks like on top of this refactor it's really as easy as #65.

PR #72, opened 2022-10-18
Add Proj.bounds function

simonbyrne:
Now we've upgraded the library, we can expose this interface.

See discussion in https://github.com/MakieOrg/GeoMakie.jl/pull/101

### DIFF
--- a/src/coord.jl
+++ b/src/coord.jl
@@ -180,6 +180,24 @@ function (trans::Transformation)(coord)::NTuple234
     end
 end
 
+"""
+    Proj.bounds(trans::Transformation, (xmin, xmax), (ymin,ymax); densify_pts=21) -> ((bxmin, bxmax), (bymin, bymax))
+
+Transform boundary densifying the edges to account for nonlinear transformations along
+these edges and extracting the outermost bounds. Returns a tuple of tuples of the bounding
+rectangle.
+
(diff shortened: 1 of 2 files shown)


rafaqz:
There is also https://github.com/rafaqz/Extents.jl that we use in GeoInterface.jl and DimensionslData.jl/Rasters.jl for this.

It's basically bounds with a type wrapper and explicit dimension name and order

simonbyrne:
@rafaqz I'm not quite sure I understand: are you suggesting we import Extent into Proj, or import Proj into Extent?

rafaqz:
import Extent

It's a tiny package just to have a shared wrapper for extents/bounding boxes.

Maybe I misunderstand what your `bounds` method is here

visr (approved):
This looks good to me, thanks for adding it!

visr:
@rafaqz it probably makes sense to also support Extents, but that can be done in a separate PR, right? Perhaps by adding a method to Proj.bounds that returns an Extent when that is the input? Or would this definition be in the way?

simonbyrne:
Probably best to do that as another PR?

I've opened https://github.com/MakieOrg/GeoMakie.jl/pull/132, which uses this.

rafaqz:
Yeah it can totally be in another PR, I was just pointing out we mostly are using `Extent` for passing around this information now in JuliaGeo.

simonbyrne:
can we merge this and tag a new release?

PR #73, opened 2022-11-04
Add Transformation constructor for pipeline and fix show for it.

### LINKED ISSUE #69: Create Transformation with a single pipeline projection
Now to setup a pipeline, we need to handle the C API:

```julia
julia> pipe = Proj.proj_create("+proj=pipeline +step +proj=unitconvert +xy_in=deg +z_in=m +xy_out=rad +z_out=m +step +inv +proj=longlat +a=6378136.3 +rf=298.257 +e=0.08181922146 +step +proj=cart +a=6378136.3 +rf=298.257 +step +inv +proj=cart +ellps=WGS84 +step +proj=unitconvert +xy_in=rad +z_in=m +xy_out=deg +z_out=m +step +proj=axisswap +order=2,1")
Ptr{Nothing} @0x00000001254061d0

julia> Proj.proj_trans(pipe, Proj.PJ_FWD, (0,0,0))
4-element Proj.Coord:
  0.0
  0.0
 -0.7000000001862645
 Inf
```

evetion:
Fixes #69

### DIFF
--- a/src/coord.jl
+++ b/src/coord.jl
@@ -1,5 +1,6 @@
 """
     Transformation(source_crs, target_crs; always_xy=false, direction = PJ_FWD, area=C_NULL, ctx=C_NULL)
+    Transformation(pipeline; always_xy=false, direction = PJ_FWD, area=C_NULL, ctx=C_NULL)
 
 Create a Transformation that is a pipeline between two known coordinate reference systems.
 Transformation implements the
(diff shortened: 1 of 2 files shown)


src/coord.jl
```diff
@@ -83,6 +86,17 @@ function Transformation(
     return Transformation(pj, direction)
 end
 
+function Transformation(
+    pipeline::AbstractString,
```
visr, line 90:
```suggestion
    pipeline::AbstractString;
```

src/coord.jl
```diff
@@ -83,6 +86,17 @@ function Transformation(
...
+    always_xy::Bool = false,
+    direction::PJ_DIRECTION = PJ_FWD,
```
visr, line 92: Why is `area::Ptr{PJ_AREA} = C_NULL` not included like in the other constructors? Or does that not apply to pipelines?

visr (changes requested):
This is great to have, thanks. I also like the Base.show improvements.

1. The new single arg AbstractString Transformation constructor is supposed to represent a pipeline. Can we check in the constructor after `proj_create` if this is really the case, and throw an error if not? Or is there no PROJ API for that? I'm worried about people creating a `Proj.Transformation("EPSG:28992")`, which (I guess) doesn't make much sense or will error.
2. Could you add a test for using the new constructor, including one that throws in the example above?

src/coord.jl
evetion: Area is used to restrict the search for a pipeline between two CRS. Here we directly construct a pipeline, so no search is involved.

visr (approved):

PR #74, opened 2022-11-04
Add CRS constructor.

evetion:


src/crs.jl
```diff
@@ -0,0 +1,83 @@
...
+- a WKT string,
+- an object code (like “EPSG:4326”, “urn:ogc:def:crs:EPSG::4326”, “urn:ogc:def:coordinateOperation:EPSG::1671”),
+- an Object name. e.g “WGS 84”, “WGS 84 / UTM zone 31N”. In that case as uniqueness is not guaranteed, heuristics are applied to determine the appropriate best match.
+- a OGC URN combining references for compound coordinate reference systems (e.g “urn:ogc:def:crs,crs:EPSG::2393,crs:EPSG::5717” or custom abbreviated syntax “EPSG:2393+5717”),
+- a OGC URN combining references for concatenated operations (e.g. “urn:ogc:def:coordinateOperation,coordinateOperation:EPSG::3895,coordinateOperation:EPSG::1618”)
+- a PROJJSON string. The jsonschema is at https://proj.org/schemas/v0.4/projjson.schema.json
+- a compound CRS made from two object names separated with “ + “. e.g. “WGS 84 + EGM96 height”
+- a GeoFormatTypes CoordinateReferenceSystemFormat such as EPSG or ProjString
+"""
+mutable struct CRS
```
visr, line 15: Would this help with anything?

```suggestion
mutable struct CRS <: GFT.CoordinateReferenceSystemFormat
```

test/libproj.jl
```diff
@@ -364,3 +365,29 @@ end
     # restore setting as outside the testset
     Proj.enable_network!(as_before)
 end
+
+@testset "CRS" begin
+    gftcrs = GFT.EPSG(4326)
+    crs = Proj.CRS(gftcrs)
```
visr, line 371: Could you add a test for the `Proj.CRS(::AbstractString)` constructor as well?

src/crs.jl
```diff
@@ -0,0 +1,83 @@
...
+        end
+        return crs
+    end
+end
+
+function CRS(
+    crs::AbstractString,
+    ctx::Ptr{PJ_CONTEXT}=C_NULL
+)
+    crs = proj_create(crs, ctx)
```
visr, line 30: Similarly to my comment in https://github.com/JuliaGeo/Proj.jl/pull/73#pullrequestreview-1169595117, should we check here if what we have created is indeed a CRS, and not say a pipeline? And test that?

src/crs.jl
```diff
@@ -0,0 +1,83 @@
...
+function is_geographic(crs::CRS)
+    proj_get_type(crs.pj) in (
+        PJ_TYPE_GEOGRAPHIC_CRS,
+        PJ_TYPE_GEOGRAPHIC_2D_CRS,
+        PJ_TYPE_GEOGRAPHIC_3D_CRS)
+end
+
+function is_projected(crs::CRS)
+    proj_get_type(crs.pj) == PJ_TYPE_PROJECTED_CRS
+end
```
visr, line 63: Do you know if this covers everything? What about PJ_TYPE_TEMPORAL_CRS or PJ_TYPE_COMPOUND_CRS? Could they also be geographic or projected? If we can't be certain, perhaps we should just stick to the PROJ API and add `proj_get_type(::CRS)` for slightly more convenience.

visr (changes requested):
Thanks for working on this. I added some comments.

src/crs.jl
evetion: I don't think so, no.

evetion:
@rafaqz This PR includes a method to convert between GFT instances, like ArchGDAL.

evetion:
Not sure why the CRS is suddenly failing, like the PROJ version changed in between. @visr, any thoughts?

visr:
That's your own show change from #73 ;)

Fixed in 29e341568db021f7c484ca65830e9fe23f4bc915

The new show is better, it shows explicitly it is a no-op (old then new):
```
Transformation
    source: WGS 84
    target: WGS 84
    direction: forward

Transformation noop
    description: Null geographic offset from WGS 84 to WGS 84
    definition: proj=noop ellps=GRS80
    direction: forward
```

evetion:
LGTM

visr:
It would be better to directly do this in the style of https://github.com/yeesian/ArchGDAL.jl/pull/349 and https://github.com/JuliaGeo/LibGEOS.jl/pull/149, since this adds a bunch of methods like `proj_get_id_code(crs::CRS) = proj_get_id_code(crs.pj)` that would soon be removed again.

evetion:
The style being 
`Base.unsafe_convert(::Type{Ptr{Cvoid}}, x::CRS) = x.ptr`
?

visr:
Yes indeed, then we don't need to take out the pointer ourselves, which is less code and safer for GC.

visr (approved):

Issue #88, opened 2023-07-20
Would be good to add a to_epsg implimentation 


alex-s-gardner:
robustly retrieving the epsg from a crs requires an implementation of pyproj's [to_epsg](https://github.com/pyproj4/pyproj/blob/1452ba404be58c14a6b64d4551c320022f5aafcf/pyproj/_crs.pyx#L2746C1-L2747C5)

visr:
Isn't this already present?

https://github.com/JuliaGeo/Proj.jl/blob/d88918ca72426245bcf52889854ad2aa08d1e7c8/src/crs.jl#L129-L133

Although it looks like this doesn't use `proj_get_id_auth_name` to check if it is EPSG, so that's a bug.

alex-s-gardner:
I believe this is a very limited version of [to_epsg](https://github.com/pyproj4/pyproj/blob/1452ba404be58c14a6b64d4551c320022f5aafcf/pyproj/_crs.pyx#L2746C1-L2747C5) as it can only find an EPSG code if the Authority and code is contained in the CRS. For example it can't infer the EPSG code from a ProjString or wkt that does not include the authority. This is the same limitation as my [ArchGDAL]( PRhttps://github.com/yeesian/ArchGDAL.jl/pull/391). `to_epsg` has a bunch of logic to try and infer the EPSG and also returns a confidence level that it found the right code.

alex-s-gardner:
I think we'll need to use [Proj.proj_identify](https://github.com/JuliaGeo/Proj.jl/blob/0e32ea7901ca5a1664d3a607dd21f8972d558367/src/libproj.jl#L1350)

alex-s-gardner:
Can someone provide an example of how to use [Proj.proj_identify](https://github.com/JuliaGeo/Proj.jl/blob/0e32ea7901ca5a1664d3a607dd21f8972d558367/src/libproj.jl#L1350), I can't seem to figure out how to assign and retrieve the C objects.

alex-s-gardner:
Should we introduce this into proj as `to_epsg()` then use that for [`GFT.EPSG(crs::CRS)`](https://github.com/JuliaGeo/Proj.jl/blob/d88918ca72426245bcf52889854ad2aa08d1e7c8/src/crs.jl#L129-L133)?

visr:
Yeah perhaps we can create a `Proj.identify` that is a nicely wrapped version of the C API `Proj.proj_identify`.

[...]

Issue #87, opened 2023-07-14
update for PROJ 9.2.1


visr:
PROJ 9.2.1 is already picked up by this package through [PROJ_jll v900.200.100](https://github.com/JuliaBinaryWrappers/PROJ_jll.jl/releases/tag/PROJ-v900.200.100%2B0).

- [ ] update tests
- [ ] re-run wrapper

visr:
The two remaining test failures can be fixed by enabling network access for those transformations. Otherwise we get these separate errors:

```
hgridshift: could not find required grid(s)
vgridshift: could not find required grid(s)

using Proj
Proj.enable_network!(false)
trans2 = Proj.Transformation("EPSG:32632", "EPSG:2027", always_xy = true) # fails, passed before, passes with network
```

I suspect this is related to the new ONLY_BEST option. This option looks like a great addition, but as I understand the PR it shouldn't break anything since it defaults to off. So I don't understand why it would affect our tests. This need some more investigation.

https://proj.org/en/9.2/news.html#id20
https://github.com/OSGeo/PROJ/pull/3535

Also perhaps we should add some support for the option just like pyproj:

https://github.com/pyproj4/pyproj/pull/1236

evetion:
I don't mind this breaking change, it should be the default imho.

PR #92, opened 2023-08-04
Proj.identify

alex-s-gardner:
This addresses the first part of #88

### DIFF
--- a/src/crs.jl
+++ b/src/crs.jl
@@ -137,3 +137,53 @@ Base.convert(::Type{CRS}, crs::GFT.CoordinateReferenceSystemFormat) = CRS(crs)
 
 # Maybe enable later, based on https://github.com/JuliaGeo/GeoFormatTypes.jl/issues/21
 # Base.convert(T::Type{<:GFT.CoordinateReferenceSystemFormat}, crs::GFT.CoordinateReferenceSystemFormat) = T(CRS(crs))
+
+
+
+
+"""
+    identify(obj; authority = nothing, options = nothing)
+
+Returns the a list of matching reference CRS or nothing, and the confidence values (0-100) 
+or nothing. This is a high level interface to proj_identify.
+
+Inputs:
+   obj -- Object of type CRS. Must not be NULL
+   authority [nothing] -- Authority name, or NULL for all authorities (e.g. "EPSG")
+   allmatches [false] -- if true return all matches, if false return top match
+   options -- Placeholder for future options. Should be set to nothing.
+"""
+function identify(obj; authority=nothing, allmatches=false, options=nothing)
+
+    out_confidence = Ref(Ptr{Cint}(C_NULL))
+    if isnothing(authority)
+        #set authority to C_NULL
+        authority = C_NULL
+    end
+
+    if isnothing(options)
+        #set authority to C_NULL
+        options = C_NULL
+    end
+
+    pj_list = Proj.proj_identify(obj, authority, out_confidence, options)
+
+    # was a match found?
+    if pj_list == C_NULL
+        crs_ref = nothing
+        confidence = nothing
+    else
+        if allmatches
+            # return all results
+            cnt = Proj.proj_list_get_count(pj_list)
+            crs_ref = [Proj.CRS(Proj.proj_list_get(pj_list, i)) for i = 0:(cnt-1)]
+            confidence = [unsafe_load(out_confidence[], i) for i = 1:cnt]
+        else
+            # only return top results
+            crs_ref = Proj.CRS(Proj.proj_list_get(pj_list, 0))
+            confidence = unsafe_load(out_confidence[], 1)
+        end
+    end
+
+    return crs_ref, confidence
+end
--- a/src/crs.jl
+++ b/src/crs.jl
@@ -1,7 +1,7 @@
 """
     CRS(crs)
 
-Create a CRS. `crs` can be:
+Create a coordinate reference system. `crs` can be:
 - a proj-string,
 - a WKT string,
 - an object code (like "EPSG:4326", "urn:ogc:def:crs:EPSG::4326", "urn:ogc:def:coordinateOperation:EPSG::1671"),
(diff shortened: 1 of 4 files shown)


alex-s-gardner:
FYI: It appears that failed tests are unrelated to this PR

alex-s-gardner:
I don't think libproj.jl is the right place to put `Proj.identify`. I think it's autogenerated. Looking through the codebase I don't see a great place to put it. Should we create a new file `utilities.jl` and put `Proj.identify` in there?

visr:
Indeed `libproj.jl` is generated so it cannot go there. I'd say the existing `crs.jl` is a fine place.

Regarding the failing tests, indeed it is a known issue, see #87.

src/crs.jl
```diff
@@ -137,3 +137,53 @@ Base.convert(::Type{CRS}, crs::GFT.CoordinateReferenceSystemFormat) = CRS(crs)
 
 # Maybe enable later, based on https://github.com/JuliaGeo/GeoFormatTypes.jl/issues/21
 # Base.convert(T::Type{<:GFT.CoordinateReferenceSystemFormat}, crs::GFT.CoordinateReferenceSystemFormat) = T(CRS(crs))
+
+
+
```
visr, line 142:
```suggestion
```
Removing extra blank lines

src/crs.jl
```diff
@@ -137,3 +137,53 @@ Base.convert(::Type{CRS}, crs::GFT.CoordinateReferenceSystemFormat) = CRS(crs)
...
+    identify(obj; authority = nothing, options = nothing)
+
+Returns the a list of matching reference CRS or nothing, and the confidence values (0-100) 
+or nothing. This is a high level interface to proj_identify.
+
+Inputs:
+   obj -- Object of type CRS. Must not be NULL
+   authority [nothing] -- Authority name, or NULL for all authorities (e.g. "EPSG")
+   allmatches [false] -- if true return all matches, if false return top match
+   options -- Placeholder for future options. Should be set to nothing.
```
visr, line 154: Shall we leave out the options placeholder while it's not used?

src/crs.jl
```diff
@@ -137,3 +137,53 @@ Base.convert(::Type{CRS}, crs::GFT.CoordinateReferenceSystemFormat) = CRS(crs)
...
+
+"""
+    identify(obj; authority = nothing, options = nothing)
+
+Returns the a list of matching reference CRS or nothing, and the confidence values (0-100) 
+or nothing. This is a high level interface to proj_identify.
+
+Inputs:
+   obj -- Object of type CRS. Must not be NULL
+   authority [nothing] -- Authority name, or NULL for all authorities (e.g. "EPSG")
```
visr, line 152: In general I'd like to stay close to the PROJ API unless there is good reason to deviate. Shall we just keep calling it `auth_name`?

src/crs.jl
```diff
@@ -137,3 +137,53 @@ Base.convert(::Type{CRS}, crs::GFT.CoordinateReferenceSystemFormat) = CRS(crs)
...
+
+
+"""
+    identify(obj; authority = nothing, options = nothing)
+
+Returns the a list of matching reference CRS or nothing, and the confidence values (0-100) 
+or nothing. This is a high level interface to proj_identify.
+
+Inputs:
+   obj -- Object of type CRS. Must not be NULL
```
visr, line 151: Is there a reason not to require `obj::CRS`?

src/crs.jl
```diff
@@ -137,3 +137,53 @@ Base.convert(::Type{CRS}, crs::GFT.CoordinateReferenceSystemFormat) = CRS(crs)
...
 # Base.convert(T::Type{<:GFT.CoordinateReferenceSystemFormat}, crs::GFT.CoordinateReferenceSystemFormat) = T(CRS(crs))
+
+
+
+
+"""
+    identify(obj; authority = nothing, options = nothing)
+
+Returns the a list of matching reference CRS or nothing, and the confidence values (0-100) 
+or nothing. This is a high level interface to proj_identify.
```
visr, line 148: This reads like it could return `nothing`, but that is not the case right? If there is no match they are just empty lists or not?

We can probably leave out the sentence about high level interface, since anything outside of libproj.jl basically is that.

src/crs.jl
```diff
@@ -137,3 +137,53 @@ Base.convert(::Type{CRS}, crs::GFT.CoordinateReferenceSystemFormat) = CRS(crs)
...
+    end
+
+    pj_list = Proj.proj_identify(obj, authority, out_confidence, options)
+
+    # was a match found?
+    if pj_list == C_NULL
+        crs_ref = nothing
+        confidence = nothing
+    else
+        if allmatches
```
visr, line 176: I think I'd prefer not adding `allmatches`, and keep it close to the PROJ API. The results are ranked in terms of confidence so it's easy to grab the first.

src/crs.jl
```diff
@@ -137,3 +137,53 @@ Base.convert(::Type{CRS}, crs::GFT.CoordinateReferenceSystemFormat) = CRS(crs)
...
+   obj -- Object of type CRS. Must not be NULL
+   authority [nothing] -- Authority name, or NULL for all authorities (e.g. "EPSG")
+   allmatches [false] -- if true return all matches, if false return top match
+   options -- Placeholder for future options. Should be set to nothing.
+"""
+function identify(obj; authority=nothing, allmatches=false, options=nothing)
+
+    out_confidence = Ref(Ptr{Cint}(C_NULL))
+    if isnothing(authority)
+        #set authority to C_NULL
```
visr, line 160: I think that's sufficiently clear from `authority = C_NULL` :)

src/crs.jl
```diff
@@ -137,3 +137,53 @@ Base.convert(::Type{CRS}, crs::GFT.CoordinateReferenceSystemFormat) = CRS(crs)
...
+            crs_ref = [Proj.CRS(Proj.proj_list_get(pj_list, i)) for i = 0:(cnt-1)]
+            confidence = [unsafe_load(out_confidence[], i) for i = 1:cnt]
+        else
+            # only return top results
+            crs_ref = Proj.CRS(Proj.proj_list_get(pj_list, 0))
+            confidence = unsafe_load(out_confidence[], 1)
+        end
+    end
+
+    return crs_ref, confidence
```
visr, line 188: In pyproj they return a list of `AuthorityMatchInfo`. I like returning a single list since it makes it clear that `crs_ref[i]` and `confidence[i]` belong together. How about we return a `Vector` of NamedTuple, with the NamedTuple being of the type `@NamedTuple{crs::CRS, confidence::Int}`?

visr (commented):
Sorry for the wait! Left some comments, curious what you think.

src/crs.jl
```diff
@@ -137,3 +137,53 @@ Base.convert(::Type{CRS}, crs::GFT.CoordinateReferenceSystemFormat) = CRS(crs)
...
+
+    if isnothing(options)
+        #set authority to C_NULL
+        options = C_NULL
+    end
+
+    pj_list = Proj.proj_identify(obj, authority, out_confidence, options)
+
+    # was a match found?
+    if pj_list == C_NULL
```
visr, line 172: Do you get this on error or also on no match found? If it's only the former it's better to throw an error here. If the latter, let's return an empty vector of the same type as normal.

src/crs.jl
```diff
@@ -137,3 +137,53 @@ Base.convert(::Type{CRS}, crs::GFT.CoordinateReferenceSystemFormat) = CRS(crs)
...
+            cnt = Proj.proj_list_get_count(pj_list)
+            crs_ref = [Proj.CRS(Proj.proj_list_get(pj_list, i)) for i = 0:(cnt-1)]
+            confidence = [unsafe_load(out_confidence[], i) for i = 1:cnt]
+        else
+            # only return top results
+            crs_ref = Proj.CRS(Proj.proj_list_get(pj_list, 0))
+            confidence = unsafe_load(out_confidence[], 1)
+        end
+    end
+
```
visr, line 187: `out_confidence` should be freed with `proj_int_list_destroy`

src/crs.jl
alex-s-gardner: sounds good.. updated

src/crs.jl
alex-s-gardner: empty list now returned, remove mention of high level interface

src/crs.jl
alex-s-gardner: sounds good.. update made

src/crs.jl
alex-s-gardner: Proj.proj_identify returns C_NULL if not match is found. We now return an empty list

src/crs.jl
alex-s-gardner: allmatches now removed

alex-s-gardner:
@visr I've adopted all of your suggestions, thanks. One question I have is should we always return a `vector` of NamedTuples even if there is no crs found or only a single crs found? My thought was that the output should be a consistent type so I always return a vector.

visr (approved):

visr:
> One question I have is should we always return a vector of NamedTuples even if there is no crs found or only a single crs found?

Yeah I agree.

I pushed a few final things in e87b0d15acdceb95aeb85a0f9dd6a53f6e9f7efd. I went back to freeing the confidence, since that is what the docs tell us, and the CRS pointers, we still use and pass back to the user. Still I couldn't trigger a segfault or anything accessing the original confidence values, so not sure if it is being cleaned up properly.

PR #107, opened 2024-05-24
Fix `geod_inverse` argument order

### LINKED ISSUE #105: Invalid order of arguments passed to geod_inverse()
The definition of the function Proj.geod_inverse() in geod.jl is as follows.

```
function geod_inverse(g::geod_geodesic, lat1::Real, lon1::Real, lat2::Real, lon2::Real)
    s12 = Ref{Cdouble}(NaN)
    azi1 = Ref{Cdouble}(NaN)
    azi2 = Ref{Cdouble}(NaN)

    geod_inverse(Ref(g), lat1, lat2, lon1, lon2, s12, azi1, azi2)

    return s12[], azi1[], azi2[]
end
```

However, I think the correct order of arguments passed to the (inner) geod_inverse() function should be

```
    geod_inverse(Ref(g), lat1, lon1, lat2, lon2, s12, azi1, azi2)
```

visr:
Fixes #105

### DIFF
--- a/src/geod.jl
+++ b/src/geod.jl
@@ -34,7 +34,7 @@ function geod_inverse(g::geod_geodesic, lat1::Real, lon1::Real, lat2::Real, lon2
     azi1 = Ref{Cdouble}(NaN)
     azi2 = Ref{Cdouble}(NaN)
 
-    geod_inverse(Ref(g), lat1, lat2, lon1, lon2, s12, azi1, azi2)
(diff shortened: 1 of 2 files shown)


asinghvi17:
Ah, I just ran into this error I think.  Thanks for the PR!  

Testing the patch reveals the correct behaviour.

asinghvi17 (approved):

PR #108, opened 2024-05-24
Wrap and require PROJ 9.5

visr:
Forgot this in #106. The newly wrapped functions have to be present in the PROJ_jll binaries that come with this package.

Also updates Aqua to the latest version, which requires compat entries for test dependencies.


visr:
We should hold off since the PROJ 9.4 build has been yanked, see https://github.com/JuliaPackaging/Yggdrasil/issues/8780.

visr:
Revived this for PROJ 9.5, which was built in https://github.com/JuliaPackaging/Yggdrasil/pull/9548.
This should be the last minor PROJ release that needs to be released as a breaking JLL.

PR #116, opened 2025-04-13
Run CI on pre releases instead of nightly

visr:
Currently our CI badge on the readme is always red due to failing tests on nightly, and in https://github.com/JuliaGeo/Proj.jl/actions/runs/14427702016 I also see this warning:

> x64 arch has been requested on a macOS runner that has an arm64 (Apple Silicon) architecture. You may have meant to use the "aarch64" arch instead (or left it unspecified for the correct default).

So this removes the architecture, and uses 1.12-beta1 rather than the master branch.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -25,6 +25,8 @@ jobs:
           - ubuntu-latest
           - macOS-latest
           - windows-latest
+        arch:
+          - default
     steps:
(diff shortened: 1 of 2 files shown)


visr:
Hmm Julia 1.6 doesn't have that macOS arch. I guess we might as well require Julia 1.10?
For the Julia 1.12 issue I made #117.

PR #118, opened 2025-05-17
add GeoFormatyTypes convert method piracy on ProjJSON

rafaqz:
This PR adds `Base.convert` to and from `GFT.ProjString`

Maybe this will cause ambiguities with ArchGDAL? I'm not sure. 

I don't love adding more piracy but what else to do?


asinghvi17:
I had a thingy for this in ArchGDAL, we can lock the version and do that whole dance

